Issue #3, opened 2017-06-05
Simple Refactoring Proposal for emmy Client


miha-plesko:
Currently emmy Client is implemented as follows:
```go
type Client struct {
    id      int32
    conn    *grpc.ClientConn
    client  *pb.ProtocolClient
    stream  pb.Protocol_RunClient
    schema  pb.SchemaType // int32
    variant pb.SchemaVariant
    handler *ClientHandler
}

func (c *Client) Pedersen(dlog *dlog.ZpDLog, val big.Int) error {
    c.handler.pedersenCommitter = commitments.NewPedersenCommitter(dlog)
    ... // c.handler.pedersenCommitter is used
}

func (c *Client) Schnorr(protocolType common.ProtocolType, dlog *dlog.ZpDLog, secret big.Int) error {
    c.handler.schnorrProver = dlogproofs.NewSchnorrProver(dlog, protocolType)
    ... // c.handler.schnorrProver is used
}
```

So example has to be started like this:
```go
genClient := getGenericClient(endpoint, example)
genClient.ExecuteProtocol(protocolParams)  // decides whether c.Pedersen() whether c.Schnorr() is run
```
What bothers me is that such an approach is hard to read. Also, when adding support for new schema, one has to modify everything: the Client struct (hook additional function), the ClientHandler struct (store additional Commiter) and the ExecuteProtocol() function (additional case).

# Proposed change
What if Client was an interface:
```go
type Client interface{
    ExecuteProtocol()
}
```
and we implement it for each schema we provide:
```go
type PedersenClient struct{}

func (c *PedersenClient) ExecuteProtocol(){
    // bouncing protocol is run
}
```
```go
type SchnorrClient struct{}

func (c *SchnorrClient) ExecuteProtocol(){
    // bouncing protocol is run
}
```
All we had to do to start emmy client then is to pick appropriate implementation at the very beginning. So example would be started like this:
```go
func PickClient(schemaName string) Client{
    switch schemaName{
    case "pedersen":
        return &PedersenClient{}
    case "schnorr":
        return &SchnorrClient{}
    }
}

client := PickClient(schemaName)
client.ExecuteProtocol()
```
*Notice how `PickClient()` return type is the interface. This means that your program, from this point on, does not care what implementation is underneath.*

[...]

PR #5, opened 2017-06-19
Refactoring of emmy clients

mancabizjak:
This is the initial attempt for refactoring of emmy clients, as agreed in [issue #3](https://github.com/xlab-si/emmy/issues/3). I used client-specific structs as well as a generic client of type `GenericClient` embedded in these structs. In some cases (Pedersen and PedersenEC) I embedded a different common struct to avoid duplication of functions. Furthermore, in cases we have something more than sigma protocol (see Schnorr and SchnorrEC), I provided seperate functions for running sigma protocol or zero knowledge proof/proof of knowledge.

However, I think there's still room for further improvements - for instance in `client.go` there's still a big `switch` block, and very similar logic in each of its `case`s.

There are two minor modifications of the crypto-backend (see files `dlogproofs/schnorr.go` and `dlogproofs/schnorr_ec.go`), so that they return a `*big.Int` instead of `nil` - otherwise we have to do this logic in client functions taking care of communication.

Signed-off-by: Manca Bizjak <manca.bizjak@xlab.si>


client/pedersen.go
```diff
@@ -7,23 +7,48 @@ import (
 	"math/big"
 )
 
-func (c *Client) Pedersen(dlog *dlog.ZpDLog, val *big.Int) error {
-	c.handler.pedersenCommitter = commitments.NewPedersenCommitter(dlog)
+type PedersenCommonClient struct {
```
miha-plesko: Is there really any benefit if we use `PedersenCommonClient`? I think it only has one method, `decommit()`, but this method is not reused by anyone, since `PedersenECClient` has it redefined [here](https://github.com/mancabizjak/emmy/blob/15211dc84867321699ab4e3c6c8bf65fdbac0362/client/pedersen_ec.go#L92). Note that I might be missing something.

client/schnorr.go
```diff
@@ -9,58 +9,105 @@ import (
...
+	}, nil
+}
 
-		c.handler.schnorrProver.PedersenReceiver.SetCommitment(commitment)
-		pedersenDecommitment, err := c.proofRandomData(false, dlog.G, secret)
-		if err != nil {
-			return err
-		}
+// Run calls the appropriate function to execute the Schnorr protocol
+func (c *SchnorrClient) Run() error {
```
miha-plesko: I like it very much how this function is now all compact and readable, job well done. Clients are now nicely isolated one from each other which make it awesomely readable so 👍

client/schnorr.go
```diff
@@ -76,26 +123,18 @@ func (c *Client) open(openMsg *pb.Message) (*big.Int, error) {
 	return new(big.Int).SetBytes(bigint.X1), nil
 }
 
-func (c *Client) proofRandomData(isFirstMsg bool, a, secret *big.Int) (*pb.PedersenDecommitment, error) {
-
-	x := c.handler.schnorrProver.GetProofRandomData(secret, a)
-	b, _ := c.handler.schnorrProver.DLog.Exponentiate(a, secret)
-
+func (c *SchnorrClient) proofRandomData(isFirstMsg bool, msg *pb.Message) (*pb.PedersenDecommitment, error) {
```
miha-plesko: Nitpicking, but did you perhaps mean `proveRandomData` instead of `proofRandomData`? Feels like it has to be a verb instead noun, but I'm not sure.

client/schnorr_ec.go
```diff
@@ -9,70 +9,110 @@ import (
...
+			Y: dlog.Curve.Params().Gy,
+		},
+	}, nil
+}
 
-	a := &common.ECGroupElement{
-		X: dlog.Curve.Params().Gx,
-		Y: dlog.Curve.Params().Gy,
+// Run calls the appropriate function to execute the Schnorr EC protocol.
+func (c *SchnorrECClient) Run() error {
```
miha-plesko: I'm thinking, would it perhaps be even better to just export `RunSigma()` and `RunZeroKnowledge()` in favor of deleting `Run()`?

miha-plesko: In case you decide to go with this option, please do the same for Schnorr too.

dlogproofs/schnorr.go
```diff
@@ -63,7 +63,7 @@ func (prover *SchnorrProver) GetProofData(challenge *big.Int) (*big.Int, *big.In
 	z.Mod(z, prover.DLog.GetOrderOfSubgroup())
 
 	if prover.protocolType != common.ZKPOK {
-		return z, nil
+		return z, new(big.Int)
```
miha-plesko: This shouldn't be part of this refactoring PR. Even if it works wrong the way it is now, such a change deserves to be reviewed in a separate PR so that you can provide short explanation in its own commit message. Please revert this from this commit and open a new PR. Merging of the second PR should be trivial as there will be no conflicts.

dlogproofs/schnorr_ec.go
```diff
@@ -74,7 +74,7 @@ func (prover *SchnorrECProver) GetProofData(challenge *big.Int) (*big.Int, *big.
 	z.Mod(z, prover.DLog.GetOrderOfSubgroup())
 
 	if prover.protocolType != common.ZKPOK {
-		return z, nil
+		return z, new(big.Int)
```
miha-plesko: Please put this in the same PR as above and revert it here.

emmy.go
```diff
@@ -133,12 +133,13 @@ func runClients(n int, concurrently bool, protocolType, protocolVariant, endpoin
...
-	genClient := getProtocolClient(protocolType, protocolVariant, endpoint)
-	if genClient == nil {
-		cLogger.Critical("Error ocurred when creating client")
+	pbType, pbVariant, err := parseSchema(protocolType, protocolVariant)
+	if err != nil {
+		cLogger.Criticalf("%v", err)
 		return
 	}
 
+	c := client.NewClient(endpoint)
```
miha-plesko: Please see my comment above. Here you should be instantiating the target client directly and invoke its method afterwards.

test/communication_test.go
```diff
@@ -43,22 +43,21 @@ func TestMain(m *testing.M) {
 
 func runTestGrpcClient(schemaType pb.SchemaType, schemaVariant pb.SchemaVariant,
 	protocolParams client.ProtocolParams) error {
-	c, err := client.NewProtocolClient(testGrpcServerEndpont, schemaType, schemaVariant)
-	if err != nil {
-		return err
-	}
-	c.ExecuteProtocol(protocolParams)
-	return nil
+	c := client.NewClient(testGrpcServerEndpont)
```
miha-plesko: Same here - instantiate target client directly and invoke its method that you want to test.

client/client.go
```diff
@@ -22,70 +18,48 @@ import (
...
-func NewProtocolClient(endpoint string, schemaType pb.SchemaType,
-	schemaVariant pb.SchemaVariant) (*Client, error) {
-	logger.Debugf("Creating client [SchemaType = %v][SchemaVariant = %v]", schemaType,
-		schemaVariant)
-
-	if err := validateSchema(schemaType, schemaVariant); err != nil {
-		return nil, err
-	}
-
+func NewGenericClient(endpoint string) (*GenericClient, error) {
```
mancabizjak: Actually, I am using this function to initialize the embedded struct. I call it when I create any protocol-specific client, for instance in `client/cspaillier.go` we have
```go
 func NewCSPaillierClient(endpoint string, pubKeyPath string, m, l *big.Int) (*CSPaillierClient, error) {
   	genericClient, err := NewGenericClient(endpoint)
    	if err != nil {
    		return nil, err
       }
    // ... initialize CSPaillierClient
}
``` 
It's similar in other protocol clients.

client/client.go
```diff
@@ -100,99 +74,114 @@ func NewProtocolClient(endpoint string, schemaType pb.SchemaType,
...
-	schemaVariant := pb.SchemaVariant_name[int32(c.variant)]
-	variant := common.ToProtocolType(c.variant)
+func NewClient(endpoint string) *Client {
+	return &Client{
+		endpoint: endpoint,
+	}
+}
 
-	logger.Infof("Starting client [%v] %v (%v)", c.id, schemaType, schemaVariant)
+func (c *Client) ExecuteProtocol(schema pb.SchemaType, variant pb.SchemaVariant,
```
mancabizjak: You're absolutely right. I moved this logic to CLI (`emmy.go`) and removed the `ExecuteProtocol` function from `client.go` altogether. In addition, now that the user of the library will be directly instantiating the desired protocol client, this removes the need for `map[string]*big.Int protocolParams`  in `client.go` (which is something that I wanted to get rid of as soon as possible anyway :+1:).

emmy.go
```diff
@@ -133,52 +135,71 @@ func runClients(n int, concurrently bool, protocolType, protocolVariant, endpoin
...
-	// Placeholder for values used to bootstrap the chosen protocol
-	// The actual (example-specific) values will be filled in
-	// depending on which protocol we are starting
-	protocolParams := client.ProtocolParams{}
+	var protocolErr error
 
 	switch protocolType {
 	case "pedersen":
-		protocolParams["commitVal"] = big.NewInt(121212121)
+		commitVal := big.NewInt(121212121)
```
miha-plesko: Is this dummy initialization still needed? If yes, we should probably move it into `NewPedersenClient` function. IMO such an initialization should never be needed so please double check if it can be avoided. Again, I might be missing something.

client/client.go
```diff
@@ -145,54 +52,52 @@ func (c *Client) close() error {
 	return nil
 }
 
-func (c *Client) ExecuteProtocol(params ProtocolParams) {
-	schemaType := pb.SchemaType_name[int32(c.schema)]
-	schemaVariant := pb.SchemaVariant_name[int32(c.variant)]
-	variant := common.ToProtocolType(c.variant)
+func NewGenericClient(endpoint string) (*genericClient, error) {
```
mancabizjak: Indeed. I removed export of generic client initialization function (now it's named `newGenericClient` as you suggested).

Personally, I prefer to keep the `struct`s on top before initializer functions. But indeed I wasn't being 100% consistent. I used your suggested template and moved auxiliary function `validateSchema` in `pedersen.go` to the bottom. I also modified `client.go` to follow this template.

client/cspaillier.go
```diff
@@ -6,46 +6,70 @@ import (
...
-	if err = c.openCSPaillier(m, u, e, v, label); err != nil {
-		return false, err
+func (c *CSPaillierClient) Run() error {
+	u, e, v, _ := c.encryptor.Encrypt(c.m, c.label)
+	if err := c.openCSPaillier(u, e, v); err != nil {
+		return err
 	}
 
-	challenge, err := c.cspaillierProveRandomData(u, e, label)
+	challenge, err := c.cspaillierProveRandomData(u, e)
```
mancabizjak: :smiley: :smiley: My bad. I wrote this on my own before @miha-stopar told me it's not meant to be interpreted this way, and then I forgot about it. I'm changing it for the sake of consistency. 

Actually, I just noticed that I forgot to rename all the functions hooked to `CSPaillierClient`! Again, for the sake of consistency (because the naming convention of functions hooked to other protocol clients), I'm applying these changes:
* Rename `openCSPaillier(...)` -> `open(...)`
* Rename `cspaillierProveRandomData(...)` -> `getProofRandomData(...)`
* Rename `cspaillierProveData(...)` -> `getProofData(...)`

client/pedersen.go
```diff
@@ -7,23 +7,48 @@ import (
 	"math/big"
 )
 
-func (c *Client) Pedersen(dlog *dlog.ZpDLog, val *big.Int) error {
-	c.handler.pedersenCommitter = commitments.NewPedersenCommitter(dlog)
+type pedersenCommonClient struct {
```
mancabizjak: Yes, definitely. Before refactoring we had a file `commitments.go` where we kept common functionality for both Pedersen and Pedersen_EC, but until now the common structs and functions were kept in `pedersen.go`. This is now fixed and we again have a seperate file for common functionality called `pedersen_common.go`

client/pedersen_ec.go
```diff
@@ -32,10 +44,47 @@ func (c *Client) PedersenEC(val *big.Int) error {
...
+		Content:  &pb.Message_Empty{&pb.EmptyMsg{}},
+	}
+	if err := c.send(initMsg); err != nil {
+		return nil, err
+	}
+
+	resp, err := c.receive()
+	if err != nil {
+		return nil, err
+	}
```
mancabizjak: I thought about this even before refactoring, but somehow I didn't remember implementing it now. I like your suggestion, and I really like the resulting simplifications in protocol client's function that do this bouncing! I decided not to name the new bouncing function `bounce`, but `getResponseTo`. So this is what we have in our functions now:

```go
message := &pb.Message{...} // form the message
response, err := c.getResponseTo(message) // instead of c.recv()
if err != nil {
	return false, err
}
// do something with response
```

client/pedersen_ec.go
```diff
@@ -32,10 +44,47 @@ func (c *Client) PedersenEC(val *big.Int) error {
...
+		},
+	}
+
+	if err := c.send(commitmentMsg); err != nil {
+		return err
+	}
+
+	if _, err := c.receive(); err != nil {
+		return err
+	}
```
mancabizjak: Done (in all clients).

client/cspaillier.go
```diff
@@ -6,46 +6,70 @@ import (
...
+		genericClient: *genericClient,
+		encryptor:     encryptor,
+		m:             m,
+		label:         l,
+	}, nil
+}
 
-	if err = c.openCSPaillier(m, u, e, v, label); err != nil {
-		return false, err
+func (c *CSPaillierClient) Run() error {
```
mancabizjak: Done. All exported functions in client package are now (briefly) documented - this includes initialization functions and `Run()` functions.

I intentionally didn't document what all the parameters for clients' initialization functions are. I will ask @miha-stopar to help me with this some time in the future to avoid silly mistakes.

emmy.go
miha-stopar: @mancabizjak, @miha-plesko yes, it is parameter sent from outside by user of the library.

client/schnorr_ec.go
```diff
@@ -9,70 +9,113 @@ import (
...
+		Schema:        pb.SchemaType_SCHNORR_EC,
+		SchemaVariant: c.variant,
+		Content:       &pb.Message_EcGroupElement{ecge},
+	}
 
-	openMsg.Content = &pb.Message_EcGroupElement{ecge}
-
-	err := c.send(openMsg)
-	if err != nil {
+	if err := c.send(openMsg); err != nil {
```
miha-plesko: I think we could make use of `getResponseTo` here.

client/schnorr_ec.go
```diff
@@ -85,27 +128,29 @@ func (c *Client) openEC(openMsg *pb.Message) (*common.ECGroupElement, error) {
...
-			X: common.ToPbECGroupElement(x),
-			A: common.ToPbECGroupElement(a),
-			B: common.ToPbECGroupElement(b),
-		},
+		&pRandomData,
 	}
-
-	err := c.send(req)
-	if err != nil {
+	if err := c.send(req); err != nil {
```
miha-plesko: Please make use of `getResponseTo` here as well.

client/schnorr_ec.go
```diff
@@ -117,32 +162,26 @@ func (c *Client) proofRandomDataEC(isFirstMsg bool, a *common.ECGroupElement, se
...
+			&pb.SchnorrProofData{
+				Z:        z.Bytes(),
+				Trapdoor: trapdoor.Bytes(),
+			},
 		},
 	}
-
-	err := c.send(msg)
-	if err != nil {
+	if err := c.send(msg); err != nil {
```
miha-plesko: Same here, please shorten with  `getResponseTo`.

emmy.go
miha-plesko: Ahhh, I understand now. So in deed I was missing something. Ok, all good then 👍

client/schnorr_ec.go
mancabizjak: Will do a Ctrl+F next time :smiley:  Done.

client/schnorr_ec.go
mancabizjak: Done.

client/schnorr_ec.go
mancabizjak: Done.

client/cspaillier.go
```diff
@@ -6,46 +6,73 @@ import (
...
-
-	openMsg := c.getInitialMsg()
-	openMsg.Content = &pb.Message_CsPaillierOpening{&opening}
+	openMsg := &pb.Message{
+		ClientId: c.id,
+		Schema:   pb.SchemaType_CSPAILLIER,
+		Content:  &pb.Message_CsPaillierOpening{&opening},
+	}
 
 	if err := c.send(openMsg); err != nil {
```
mancabizjak: Done.

mancabizjak:
Okay, I think the client refactoring is now complete :+1: I ran into some troubles with unit tests and was only able to provide a temporary workaround for a problematic test to pass, in order not to mess with the server-side code as a part of this PR. I hope this suffices for now.

mancabizjak:
This resolves xlab-si/emmy#3.

mancabizjak:
Great! Certainly.

mancabizjak:
OK, done.

miha-plesko:
Thank you for the contribution @mancabizjak your PR has been merged! 🎆 🍾 😄 I think it will be quite trivial to add new Clients from now on.

PR #9, opened 2017-08-04
Improve tests and implement benchmarks

mancabizjak:
This commit adds benchmarking of crypto schemes to emmy, cleans up & improves existing tests. For the purpose of easier benchmarking, all protocol clients now receive `*grpc.ClientConn` instead of a `string` holding server endpoint - connection handling was moved outside of clients. The clients, whether started concurrently or sequentially, can now send their RPCs via the same connection, which reduces some initialization overhead when running tests and benchmarks.

Several other necessarry improvements were also made in order to support a better testbed enviroment, e.g. a more intuitive way of starting emmy server and some logging options (e.g. it is now possible to discard the logs emmitted by emmy server and client, as they impact performance).

The benchmarks are implemented in a per-schema way, in the `benchmark` package. Tests are in the `test` package (which was also renamed). Both benchmarks and tests share some common functionality, specifically, definitions of functions that instantiate and run specific protocol clients for testing purposes (please see `test/common.go`).


mancabizjak:
Closing in favor of #10, #11, #12 and #13.

PR #10, opened 2017-08-07
Enable log suppression

mancabizjak:
This commit adds a TurnOff function to emmy's logging package. The function
is useful for suppressing otherwise intentionally verbose logging, which is not 
always desired (for instance, when running tests or benchmarks).

### DIFF
--- a/log/logging.go
+++ b/log/logging.go
@@ -2,6 +2,9 @@ package log
 
 import "github.com/op/go-logging"
 
+var clientModule = "client"
+var serverModule = "server"
+
 var longFormat = logging.MustStringFormatter(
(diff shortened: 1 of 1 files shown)


miha-plesko:
I don't mind merging this as it is although I'd prefer if we also call this function somewhere already. Including [this line](https://github.com/xlab-si/emmy/pull/9/files#diff-6816ffa0a2c19b48b5946e116d50392cR28) in this commit would probably make sense - what do you think?

I'm asking  because I see that you will have a package name collision between `log` and `github.com/xlab-si/emmy/log`. If you decide to resolve it here, please go ahead, otherwise let me know and we merge as it is.

miha-plesko:
@mancabizjak thanks. As discussed offline, we will suppress logs in unit tests later when we are refactoring them anyway.

PR #11, opened 2017-08-07
Improve server setup and teardown logic

mancabizjak:
Until now, the `Server` struct and corresponding `NewProtocolServer` initializer in `server/server.go` were empty, so code in `emmy.go` and `test/communication_test.go` had to take care of server initialization, starup and teardown. This commit modifies the Server struct by enclosing the` *grpc.Server`,  which allows setup and teardown directly via `Server` struct, resulting in simplified server management in tests and in `emmy.go`.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2,25 +2,93 @@ package server
 
 import (
 	"fmt"
+	"github.com/grpc-ecosystem/go-grpc-prometheus"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/xlab-si/emmy/common"
 	"github.com/xlab-si/emmy/config"
 	"github.com/xlab-si/emmy/log"
 	pb "github.com/xlab-si/emmy/protobuf"
+	"google.golang.org/grpc"
 	"io"
+	"math"
+	"net"
+	"net/http"
 	"path/filepath"
 )
 
 var _ pb.ProtocolServer = (*Server)(nil)
 
-type Server struct{}
+type Server struct {
+	grpcServer *grpc.Server
+}
 
 var logger = log.ServerLogger
 
+// NewProtocolServer initializes an instance of the Server struct and returns a pointer.
+// It performs some default configuration (tracing of gRPC communication and interceptors)
+// and registers RPC protocol server with gRPC server.
 func NewProtocolServer() *Server {
 	logger.Info("Instantiating new protocol server")
-	// At the time of instantiation, we don't yet know which handler or stream to use,
-	// therefore just return a reference to the empty struct
-	return &Server{}
+
+	// Register our generic service
+	logger.Info("Registering services")
+
+	// Allow as much concurrent streams as possible and register a gRPC stream interceptor
+	// for logging and monitoring purposes.
+	server := &Server{
+		grpcServer: grpc.NewServer(
+			grpc.MaxConcurrentStreams(math.MaxUint32),
+			grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
+		),
+	}
+
+	// Disable tracing by default, as is used for debugging purposes.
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -2,25 +2,71 @@ package server
...
+// and registers RPC protocol server with gRPC server.
 func NewProtocolServer() *Server {
 	logger.Info("Instantiating new protocol server")
-	// At the time of instantiation, we don't yet know which handler or stream to use,
-	// therefore just return a reference to the empty struct
-	return &Server{}
+
+	// Start new gRPC server and register services, while allowing
+	// as much concurrent streams as possible
+	grpc.EnableTracing = true
```
miha-plesko: Should this be set in `Start()` method?  I see following documentation for this field:
```go
// EnableTracing controls whether to trace RPCs using the golang.org/x/net/trace package.
// This should only be set before any RPCs are sent or received by this program.
var EnableTracing = true
```

Anyway, are you sure we even need the tracing? As far as I understand, it exposes
```
/debug
/requests
/debug/events
```
HTTP endpoints that can be used for debugging. But do we really consume them anywhere? I may be missing something, but I'd put set this in `Start()` method only when required by user (additional argument), by default I'd set it to `false`.

server/server.go
```diff
@@ -2,25 +2,71 @@ package server
...
+	logger.Info("Registering services")
+
+	server := &Server{
+		grpcServer: grpc.NewServer(
+			grpc.MaxConcurrentStreams(math.MaxUint32),
+			grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
+		),
+	}
+	pb.RegisterProtocolServer(server.grpcServer, server)
+	grpc_prometheus.Register(server.grpcServer) // Enable debugging
```
miha-plesko: I'm curious what kind of debugging are we enabling here? If I comment this line out all the unit tests seem to still pass. This comment is more like a question than a request to remove it since I'm probably missing something.

server/server.go
```diff
@@ -2,25 +2,71 @@ package server
...
+// Start configures and starts the protocol server at the requested port.
+func (s *Server) Start(port int) {
+	connStr := fmt.Sprintf(":%d", port)
+	listener, err := net.Listen("tcp", connStr)
+	if err != nil {
+		logger.Criticalf("Could not connect: %v", err)
+	}
+
+	http.Handle("/metrics", prometheus.Handler())
+	go http.ListenAndServe(":8881", nil)
```
miha-plesko: I think L60 and L61 are needed for debugging only, right? If yes, would you consider making them optional? I'm thinking something like
```
emmy server -> starts server without debugging
```
or
```
emmy server --debug -> starts server with all the debugging
```

test/communication_test.go
```diff
@@ -8,38 +8,21 @@ import (
...
+// the server.
 func TestMain(m *testing.M) {
-	server := setupTestGrpcServer()
+	server := server.NewProtocolServer()
+	go server.Start(7008)
 	returnCode := m.Run()
-	teardownTestGrpcServer(server)
+	server.Teardown()
 	os.Exit(returnCode)
 }
```
miha-plesko: We could make use of `log.TurnOff()` here since we didn't do it it the previous PR ;)

miha-plesko:
And the screenshot:

capture

(notice how they break)

test/communication_test.go
miha-plesko: Fijuuuh. This was only spotted because we split the big PR into smaller ones 😄

mancabizjak: I see your point :wink:

mancabizjak:
Okay, done!

miha-plesko:
Merged 🎆 🍾 🥇

PR #13, opened 2017-08-08
Create gRPC connections outside protocol clients

mancabizjak:
This commit moves creation of `*grpc.ClientConn` outside protocol clients. Instead of server `endpoint string`, clients are now passed a connection, which allows us to:
* re-use a single connection across several protocol clients
* reduce some initialization overhead when running tests (and in future benchmarks)
* avoid an issue with exhausted file descriptors in case we are starting too many connections, which occurs when starting thousands of concurrent clients on the same host

Inside main client functions executing crypto protocols, you will notice that once the protocol finishes we are not manipulating (closing) connection anymore, but gRPC streams. E.g., instead of this:

```go
if err := c.close(); err != nil { ... } // closes gRPC connection
```

we are now doing:

```go
if err := c.stream.CloseSend(); err != nil { ... } // closes client-side gRPC stream
```

Clients, unit tests and `emmy.go` CLI were all updated accordingly.

### DIFF
--- a/client/cspaillier.go
+++ b/client/cspaillier.go
@@ -3,6 +3,7 @@ package client
 import (
 	"github.com/xlab-si/emmy/encryption"
 	pb "github.com/xlab-si/emmy/protobuf"
+	"google.golang.org/grpc"
 	"math/big"
 )
 
@@ -13,8 +14,8 @@ type CSPaillierClient struct {
 }
 
 // NewCSPaillierClient returns an initialized struct of type CSPaillierClient.
-func NewCSPaillierClient(endpoint string, pubKeyPath string, m, l *big.Int) (*CSPaillierClient, error) {
-	genericClient, err := newGenericClient(endpoint)
+func NewCSPaillierClient(conn *grpc.ClientConn, pubKeyPath string, m, l *big.Int) (*CSPaillierClient, error) {
+	genericClient, err := newGenericClient(conn)
 	if err != nil {
 		return nil, err
 	}
@@ -35,6 +36,9 @@ func NewCSPaillierClient(endpoint string, pubKeyPath string, m, l *big.Int) (*CS
 // Run runs the Camenisch-Shoup sigma protocol for verifiable encryption and decryption
 // of discrete logatirhms.
 func (c *CSPaillierClient) Run() error {
+	c.openStream()
+	defer c.closeStream()
+
 	u, e, v, _ := c.encryptor.Encrypt(c.m, c.label)
 	if err := c.open(u, e, v); err != nil {
 		return err
@@ -47,10 +51,6 @@ func (c *CSPaillierClient) Run() error {
 
 	_, err = c.getProofData(challenge)
 	if err != nil {
-		return err
-	}
-
-	if err := c.close(); err != nil {
 		return err
 	}
 
--- a/client/pseudonymsys.go
+++ b/client/pseudonymsys.go
@@ -8,21 +8,25 @@ import (
 	"github.com/xlab-si/emmy/dlogproofs"
 	pb "github.com/xlab-si/emmy/protobuf"
 	"github.com/xlab-si/emmy/pseudonymsys"
+	"google.golang.org/grpc"
 	"math/big"
 )
 
 type PseudonymsysClient struct {
 	genericClient
-	endpoint string
-	dlog     *dlog.ZpDLog
+	dlog *dlog.ZpDLog
 }
 
-func NewPseudonymsysClient(endpoint string) (*PseudonymsysClient, error) {
+func NewPseudonymsysClient(conn *grpc.ClientConn) (*PseudonymsysClient, error) {
+	genericClient, err := newGenericClient(conn)
+	if err != nil {
+		return nil, err
+	}
 	dlog := config.LoadDLog("pseudonymsys")
 
 	return &PseudonymsysClient{
-		endpoint: endpoint,
-		dlog:     dlog,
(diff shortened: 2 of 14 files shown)


client/pseudonymsys_ca_ec.go
```diff
@@ -15,8 +15,8 @@ type PseudonymsysCAClientEC struct {
 	prover *dlogproofs.SchnorrECProver
 }
 
-func NewPseudonymsysCAClientEC(endpoint string) (*PseudonymsysCAClientEC, error) {
-	genericClient, err := newGenericClient(endpoint)
+func NewPseudonymsysCAClientEC(conn *grpc.ClientConn) (*PseudonymsysCAClientEC, error) {
```
miha-plesko: It would be best to enforce `<whatever>Client` naming convention. Here I see you use:
```
PseudonymsysCAClientEC
```
but I would expect something like:
```
PseudonymsysCaEcClient
```
Please consider unifying this in one of the following PRs (not in this one).

test/pseudonymsys_ec_test.go
```diff
@@ -13,7 +13,7 @@ import (
 
 func TestPseudonymsysEC(t *testing.T) {
 	dlog := dlog.NewECDLog(dlog.P256)
-	caClient, err := client.NewPseudonymsysCAClientEC(testGrpcServerEndpoint)
+	caClient, err := client.NewPseudonymsysCAClientEC(testGrpcClientConn)
```
miha-plesko: I'm not sure if we're allowed to refer global variables from other test files. There is no guarantee that communication_test.go will be executed prior pseudonymsys_ec_test.go. Also, in communication_test.go we invoke
```go
testGrpcClientConn.Close()
```
so I'm not sure how is this currently even working?

On some other project*, we resolved similar situation by using gocheck framework for unit testing instead of builtin Go tests. The framework allows you to create a struct with arbitrary fields (you could store connection there) and specify functions like
```
SetUpSuite // before running all the tests
SetUpTest //before running each test
TearDownTest // after running each test
TearDownSuite // after running all the tests
```
Do you think it would make sense to start using the gocheck also for emmy?

\* https://github.com/mikelangelo-project/capstan/blob/master/cmd/package_test.go

mancabizjak:
@miha-plesko I further improved the clients by taking into account your comments regarding opening and closing of communication streams.

Please note the updated struct:

````go
type genericClient struct {
	id             int32
	protocolClient pb.ProtocolClient
	stream         pb.Protocol_RunClient
}
````
Notice that I removed the `*grpc.ClientConn` field and added `protocolClient` and `stream` instead. We don't actually need to keep reference to `*grpc.ClientConn` since we're now handling connections outside the scope of clients. 
However, we are manipulating streams, which is why we have this new `stream` field. To create the stream outside the `newGenericClient` constructor, I also had to add a field of type `pb.ProtocolClient` which is used as a proxy to obtain the `stream`.

File: client/pseudonymsys.go [hunk 3]
```diff
@@ -32,11 +33,13 @@ func (c *PseudonymsysClient) GenerateNym(userSecret *big.Int,
 	caCertificate *pseudonymsys.CACertificate) (
 	*pseudonymsys.Pseudonym, error) {
 	// new client needs to be created in each method to implicitly call server Run method:
-	genericClient, err := newGenericClient(c.endpoint)
+	genericClient, err := newGenericClient(c.conn)
```
miha-plesko: Please move `genericClient` struct initialization from
```
GenerateNym
ObtainCredential
TransferCredential
```
functions into the `NewPseudonymsysClient` constructor . You only need to open stream in each function.

client/pseudonymsys_ec.go
```diff
@@ -28,11 +29,13 @@ func (c *PseudonymsysClientEC) GenerateNym(userSecret *big.Int,
 	caCertificate *pseudonymsys.CACertificateEC) (
 	*pseudonymsys.PseudonymEC, error) {
 	// new client needs to be created in each method to implicitly call server Run method:
-	genericClient, err := newGenericClient(c.endpoint)
+	genericClient, err := newGenericClient(c.conn)
```
miha-plesko: Same here, please move struct initialization into the `NewPseudonymsysClientEC` constructor for all the functions:
```
GenerateNym
ObtainCredential
TransferCredential
```

test/pseudonymsys_ec_test.go
miha-plesko: Oh, I didn't realize the `TestMain` is shared for all the tests in the package. All good then, thanks for teaching me this.

client/cspaillier.go
```diff
@@ -50,11 +53,7 @@ func (c *CSPaillierClient) Run() error {
 		return err
 	}
 
-	if err := c.close(); err != nil {
-		return err
-	}
-
-	return nil
+	return c.closeStream()
```
miha-plesko: As discussed offline, I suggest we use `defer` for all the clients. This way the code will become even more readable. The downside is that the possible error that comes from stream closing won't bubble up, but I'm not sure if this is a big problem.

mancabizjak:
@miha-plesko I implemented your comments regarding consistency when using `defer` with `func closeStream()`. I also moved initialization of `genericClient` outside of pseudonymsys client and pseudonymsys_ec client protocol functions. 

I squashed recent commits into one. I think we're ready to merge :fireworks: :+1:

miha-plesko:
Good job, merged 🎆 🍾 😄

PR #16, opened 2017-08-21
Add support for controlling log levels

mancabizjak:
In this commit we modify the `logging` package in order to be more conveniently used by other (mainly `client`, `server` and `test`) packages. The default log level set by logging package's `init()` function is *info*. However for tests, we're now explicitly setting the loglevel to *debug*. The log levels are standard ones, e.g. *debug*, *info*, *notice*, *error* and *critical*.

The loglevel flag was also added to emmy CLI in order to give the end user some more control. The end user interacting with the CLI app can now control the log level like this:

````bash
$ emmy example --loglevel notice  # or '-l notice'
$ emmy server start --loglevel info # we can omit this, as info is the default log level
$ emmy client --protocol schnorr -l debug
````
I didn't  update the CLI app usage instructions in the README - I can add it to this PR, or, alternatively, do in the near future in a separate PR.

In addition, appropriate changes allowing separation of what gets logged in, for instance, *info* and *debug* log levels, were applied to logging calls in `client/client.go` and `server/server.go`. Specifically, you will notice that the payloads of protobuf messages (which were making the logs a bit messy and harder to read) only get logged if we choose the *debug* log level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ The library comes with a convenient CLI for running *emmy server*, *emmy clients
 To get some more information about the theory behind zero knowledge proofs, please refer to additional documentation in the *docs* folder.
 
 #### What does emmy stand for?
-Emmy is named after [Emmy Noether](https://sl.wikipedia.org/wiki/Emmy_Noether).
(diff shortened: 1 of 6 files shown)


mancabizjak:
Okay, done :+1: . I also updated the "What does emmy stand for" portion of the README as requested.

PR #17, opened 2017-08-21
Add TLS server-side authentication

mancabizjak:
This commit adds server-side authentication with TLS in order to secure communication chanel between clients and the server. From now on, clients cannot communicate with emmy server over an insecure channel. 

In a real setting, individual clients should be able to obtain and provide the certificate of the CA that issued emmy server's certificate, and pass the local path to this file as an argument to `func GetConnection` (see `client/client.go`). However, for testing and development purposes a sample self-signed server certificate, which can be used as the root CA certificate (`test/testdata/server.pem`) and private key file (`test/testdata/server.key`) were added.

The main changes are in `client/client.go` and `server/server.go`, which are now additionally handling configuration of the TLS channel and possible errors that might occur. For instance, you will notice that `func NewProtocolServer` in `server/server.go` now returns `(*Server, error)` instead of just `*Server` - `error` is raised in case the server cannot be started due to problems in creation of TLS credentials. This error, as well as similar errors stemming from TLS credetials generation on the client side, have to be handled in `emmy.go` and `test/communication_test.go`.

In addition, the CLI in `emmy.go` was updated so that:
*  `emmy server` and `emmy example` CLI programs accept two additional flags, `--key` and `--cert` with paths to the files with server's certificate and private key. If they are not specified, the provided ones (in `test/testdata`) are used.
* Similarly, `emmy client` accepts the `--cert` flag pointing to the server certificate file.


mancabizjak:
@miha-stopar Done. Please review the updated README, then I'll squash commits and we can merge :slightly_smiling_face:

PR #18, opened 2017-08-29
Proof of partial dlog knowledge.

miha-stopar:
We have a1, b1, a2, b2 in a group G = <a1> = <a2>. We want to prove that either we know secret1 such that a1^secret1 = b1 or a2^secret2 = b2.

### DIFF
--- a/crypto/dlogproofs/partial_dlog_knowledge.go
+++ b/crypto/dlogproofs/partial_dlog_knowledge.go
@@ -1,0 +1,133 @@
+package dlogproofs
+
+import (
+	"github.com/xlab-si/emmy/crypto/common"
+	"github.com/xlab-si/emmy/crypto/dlog"
+	"github.com/xlab-si/emmy/types"
+	"math/big"
+)
+
+func ProvePartialDLogKnowledge(dlog *dlog.ZpDLog, secret1, a1, a2, b2 *big.Int) bool {
+	prover := NewPartialDLogProver(dlog)
+	verifier := NewPartialDLogVerifier(dlog)
+
+	b1, _ := prover.DLog.Exponentiate(a1, secret1)
+	triple1, triple2 := prover.GetProofRandomData(secret1, a1, b1, a2, b2)
+
+	verifier.SetProofRandomData(triple1, triple2)
+	challenge := verifier.GetChallenge()
+
+	c1, z1, c2, z2 := prover.GetProofData(challenge)
+	verified := verifier.Verify(c1, z1, c2, z2)
+	return verified
+}
+
+// Proving that it knows either secret1 such that a1^secret1 = b1 (mod p1) or
+//  secret2 such that a2^secret2 = b2 (mod p2).
+type PartialDLogProver struct {
+	DLog    *dlog.ZpDLog
+	secret1 *big.Int
+	a1      *big.Int
+	a2      *big.Int
+	r1      *big.Int
+	c2      *big.Int
+	z2      *big.Int
+	ord     int
+}
+
+func NewPartialDLogProver(dlog *dlog.ZpDLog) *PartialDLogProver {
+	return &PartialDLogProver{
+		DLog: dlog,
+	}
+}
+
+func (prover *PartialDLogProver) GetProofRandomData(secret1, a1, b1, a2,
+	b2 *big.Int) (*types.Triple, *types.Triple) {
+	prover.a1 = a1
+	prover.a2 = a2
+	prover.secret1 = secret1
+	r1 := common.GetRandomInt(prover.DLog.GetOrderOfSubgroup())
+	c2 := common.GetRandomInt(prover.DLog.GetOrderOfSubgroup())
+	z2 := common.GetRandomInt(prover.DLog.GetOrderOfSubgroup())
+	prover.r1 = r1
+	prover.c2 = c2
+	prover.z2 = z2
+	x1, _ := prover.DLog.Exponentiate(a1, r1)
+	x2, _ := prover.DLog.Exponentiate(a2, z2)
+	b2ToC2, _ := prover.DLog.Exponentiate(b2, c2)
+	b2ToC2Inv := prover.DLog.Inverse(b2ToC2)
+	x2, _ = prover.DLog.Multiply(x2, b2ToC2Inv)
+
+	// we need to make sure that the order does not reveal which secret we do know:
+	ord := common.GetRandomInt(big.NewInt(2))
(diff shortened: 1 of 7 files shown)


crypto/dlogproofs/partial_dlog_knowledge_ec.go
```diff
@@ -0,0 +1,160 @@
...
+		return false	
+	}
+
+	left1, left2 := verifier.DLog.Exponentiate(verifier.a1.X, verifier.a1.Y, z1)
+	r1, r2 := verifier.DLog.Exponentiate(verifier.b1.X, verifier.b1.Y, c1)
+	right1, right2 := verifier.DLog.Multiply(r1, r2, verifier.x1.X, verifier.x1.Y)
+	if left1.Cmp(right1) == 0 && left2.Cmp(right2) == 0 {
+	} else {
+		return false
+	}
```
mancabizjak: Please put code in L144-150 in a separate function that returns `bool`. Then, call it twice inside this function. Since it seems to me that L152-159 are a duplicate of L144-150 with the exception that L156 below returns `true` while the first block just continues. So, use the result of the first call of the newly introduced function to determine whether to continue or return, and if we continue, return the result of the second call to this newly introduced function.

Also, inside this newly introduced function, please use this logic instead of the current one in L147-150 to:

```go
return left1.Cmp(right1) == 0 && left2.Cmp(right2) == 0
```

crypto/dlogproofs/partial_dlog_knowledge.go
```diff
@@ -0,0 +1,146 @@
...
+func (verifier *PartialDLogVerifier) Verify(c1, z1, c2, z2 *big.Int) bool {
+	c := new(big.Int).Xor(c1, c2)	
+	if c.Cmp(verifier.challenge) == 0 {
+	} else {
+		return false	
+	}
+
+	left1, _ := verifier.DLog.Exponentiate(verifier.a1, z1)
+	r11, _ := verifier.DLog.Exponentiate(verifier.b1, c1)
+	right1, _ := verifier.DLog.Multiply(r11, verifier.x1)
```
mancabizjak: Put L132-135 into a separate function. Call it twice.

crypto/dlogproofs/partial_dlog_knowledge.go
```diff
@@ -0,0 +1,146 @@
...
+	
+	left2, _ := verifier.DLog.Exponentiate(verifier.a2, z2)
+	r12, _ := verifier.DLog.Exponentiate(verifier.b2, c2)
+	right2, _ := verifier.DLog.Multiply(r12, verifier.x2)
+
+	if left1.Cmp(right1) == 0 && left2.Cmp(right2) == 0 {
+		return true
+	} else {
+		return false
+	}
```
mancabizjak: Instead of L141-145, please use
```go
return left1.Cmp(right1) == 0 && left2.Cmp(right2) == 0
```

crypto/dlogproofs/partial_dlog_knowledge.go
```diff
@@ -0,0 +1,134 @@
...
+	z1.Mod(z1, prover.DLog.GetOrderOfSubgroup())
+	
+	if prover.ord == 0 {
+		return c1, z1, prover.c2, prover.z2
+	} else {
+		return prover.c2, prover.z2, c1, z1
+	}
+}
+
+type PartialDLogVerifier struct {
```
miha-stopar: I added two new types: Triple and ECTriple.

crypto/dlogproofs/partial_dlog_knowledge_ec.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (verifier *PartialECDLogVerifier) GetChallenge() *big.Int {
+	challenge := common.GetRandomInt(verifier.DLog.GetOrderOfSubgroup())
+	verifier.challenge = challenge
+	return challenge
+}
+
+func (verifier *PartialECDLogVerifier) Verify(c1, z1, c2, z2 *big.Int) bool {
+	c := new(big.Int).Xor(c1, c2)	
+	if c.Cmp(verifier.challenge) == 0 {
```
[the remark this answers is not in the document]
miha-stopar: What do you mean exactly? It needs to be changed whether c = c1 xor c2 (c is chosen by verifier, while c1 and c2 by prover).

crypto/dlogproofs/partial_dlog_knowledge.go
```diff
@@ -123,24 +112,23 @@ func (verifier *PartialDLogVerifier) GetChallenge() *big.Int {
...
+	left, _ := verifier.DLog.Exponentiate(triple.B, z) // (a, z)
+	r1, _ := verifier.DLog.Exponentiate(triple.C, challenge) // (b, challenge)
+	right, _ := verifier.DLog.Multiply(r1, triple.A) // (r1, x1)
+
+	return left.Cmp(right) == 0
+}
+
 func (verifier *PartialDLogVerifier) Verify(c1, z1, c2, z2 *big.Int) bool {
 	c := new(big.Int).Xor(c1, c2)	
 	if c.Cmp(verifier.challenge) == 0 {
```
mancabizjak: This block of code
```go
if c.Cmp(verifier.challenge) == 0 {
} else {
  	return false	
}
```
can be changed to:
```go
if c.Cmp(verifier.challenge) != 0 {
  	return false	
}
```

crypto/dlogproofs/partial_dlog_knowledge_ec.go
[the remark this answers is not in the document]
mancabizjak: Sorry, I forgot to add a clarification. So, if we look at this block of code:
```go
if c.Cmp(verifier.challenge) == 0 {
} else {
  	return false	
}
```
if basically says "if the condition evaluates to true, do nothing. Else, return false". Then, if the condition previously evaluated to true, the code continues after the if-else statement.

I wanted to suggest that we change the if-else statement like this: 
```go
if c.Cmp(verifier.challenge) != 0 {
  	return false	
}
```
since if I am not mistaken, it does what you initially wanted to do with the if-else statement, right? If I am mistaken, please ignore this comment :slightly_smiling_face: . If you decide to change it, it occurs twice (here and in `crypto/dlogproofs/partial_dlog_knowledge.go` as well).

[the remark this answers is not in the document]
miha-stopar: Agree :)

PR #22, opened 2017-09-13
Allow insecure TLS connections 

mancabizjak:
Since #17, client and server use TLS to secure channel over which they communicate. However, for development and testing purposes we sometimes want to avoid errors stemming from the server's use of self-signed cert. 

In this PR we:
* add an additional parameter to client's `GetConnection()`, which specifies whether to avoid checking of server's CA cert chain and hostname. 
* Update the calls in the test code and client CLI code.
* Update README with explanation and examples that demonstrate the purpose and usage of the `--insecure` flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,9 @@ where *commonClientFlags* control the following aspects:
 3. **Logging level**: flag *--loglevel* (shorthand *-l*), which must be one of `debug|info|notice|error|critical`. Defaults to `ìnfo`.
 4. **URI of the emmy server**: flag *--server*, defaults to *localhost:7007*.
 5. **CA certificate**: flag *--cacert*, points to a path to a certificate of the CA that issued emmy server's certificate, in PEM format. This will be used to secure communication channel with the server. Please refer to [explanation of TLS support in Emmy](#tls-support) for explanation.
+6. **Allowing insecure connections**: flag *--insecure*. In the absence of this flag clients will check emmy server's hostname and CA certificate chain. If you include this flag, none of these checks will be performed.
+  
+    > This should only be used for connecting clients to emmy development server, where self-signed certificate is used.
 
 Moreover, the *protocolSubcommand* corresponds to a concrete crypto protocol that we want to demonstrate between emmy client and emmy server. You can list valid *protocolSubcommand* values by running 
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -191,7 +194,27 @@ To control keys and certificates used for TLS, emmy CLI programs use several fla
...
+  2017/09/13 12:48:47 [client] 12:48:47.232 GetConnection ▶ INFO 001 Getting the connection
+  Cannot connect to gRPC server: Could not connect to server 10.12.13.45:7007 (x509: cannot validate certificate for 10.12.13.45 because it doesnt contain any IP SANs)
+  ```
+
+  Now let's include the *--insecure* flag, and the (insecure) connection to the server is now successfully established.
+
+  ```bash
+  $ emmy client --server 10.10.43.45:7007 --insecure schnorr
+
+  2017/09/13 12:49:14 [client] 12:49:14.961 GetConnection ▶ INFO 001 Getting the connection
```
miha-plesko: Perhaps a WARNING would be in place when `--insecure` is used?

test/communication_test.go
```diff
@@ -37,7 +37,8 @@ func TestMain(m *testing.M) {
 	go server.Start(7008)
 
 	// Establish a connection to previously started server
-	testGrpcClientConn, err = client.GetConnection(testGrpcServerEndpoint, "testdata/server.pem")
+	testGrpcClientConn, err = client.GetConnection(testGrpcServerEndpoint,
+		"testdata/server.pem", true)
```
miha-plesko: Set `insecure` to `false` here since you're passing the cert anyway.

miha-plesko: Also, could you add another test where you don't specify the cert (use empty string), and set `insecure` to `true`?

cli/flags.go
```diff
@@ -40,6 +40,12 @@ var caCertFlag = cli.StringFlag{
 	Usage: "`PATH` to certificate file of the CA that issued emmy server's certificate",
 }
 
+// concurrencyFlag indicates whether to run clients concurrently or not.
```
miha-plesko: Whoooops 😄

README.md
mancabizjak: Indeed. Done.

cli/flags.go
mancabizjak: Fixed :smile:

test/communication_test.go
mancabizjak: Done and done.

client/client.go
```diff
@@ -22,12 +21,18 @@ func SetLogLevel(level string) error {
...
 // is able to multiplex several RPCs on the same connection, thus reducing the overhead
-func GetConnection(serverEndpoint, caCert string) (*grpc.ClientConn, error) {
+func GetConnection(serverEndpoint, caCert string, insecure bool) (*grpc.ClientConn, error) {
 	logger.Info("Getting the connection")
 	timeoutSec := config.LoadTimeout()
 
+	// Notify the end user about security implications when running in insecure mode
+	if insecure {
+		logger.Warning(`######## You requested an *insecure* channel! ######## As a consequence,
+			server's identity will *NOT* be validated! Please consider using a secure connection`)
```
miha-plesko: FYI: if you use backticks in Go, then it will take all the text *literally*, including  the three tabs at the beginning of your second line. Perhaps it would be nicer to have to separate warning lines, but I'm not too picky about this.

PR #25, opened 2017-09-25
Support for multiple log backends and formats

mancabizjak:
Since #16 we had a global definition of `ClientLogger` and `ServerLogger` in the `log` package and both of these were initialized in `logging.go`'s `init()` function. In order to separate logging backends for client and server, we improved `func NewLogger()`, which is now capable of configuring more than one backend per logger. It now accepts three arguments: 
1.  `module` - a descriptive string (for instance "client" or "server"),
2. `logFilePath` - path to the file that logger will output its logs to *in addition to stdout*. If an empty string is given, file backend will not be used, e.g. logs will only be outputted to console.
3. `loglevel` - the initial loglevel of newly created logger (can be changed later with `func (logger *Logger) SetLevel()`).

Initialization of the logger used across `client` package was moved to `init()` in `client.go`. We kept this per-package logger for convenience, however the server-side logger is now hooked to a particular instance of the `Server`, as it seemed more appropriate. Logger initialization logic and calls in `client.go`, `server.go`, `cli` and `test` packages were updated accordingly. A new test was added that checks behavior in case invalid log level was passed to `func NewLogger()`. 

There are further changes, including:
* Definition of three logging formats in a file `log/formats.go`. Note that `FORMAT_SHORT` is used by default for client logger (because it doesn't need as fine-grained logs as the server), `FORMAT_LONG` is used for the server's logger when outputting to console, and `FORMAT_LONG_COLORLESS` is used for the server's logger when outputting to file (because it doesn't need color information like console output). These updated formats output logs that are much more readable than before.
* Definition of constants denoting log levels in `log/levels.go`.
* Package function `TurnOff()` was discontinued in favor of `func (logger *Logger) Disable()`.
* The log level for server and client loggers during tests was modified from `debug` to `notice`.

One more thing: this PR also removes the problem of duplicated timestamps in logs. So, instead of this:
```bash
2017/09/06 12:38:52 [server] 12:38:52.562 Start ▶ NOTI 006 Emmy server listening for connections on port 7007
```
we now have something like this:

```bash
[server][Mon 25.Sep 2017,14:11:041] Start ▶ NOTI  Emmy server listening for connections on port 7007
```

Notice how much easier it is to read now :slightly_smiling_face:


client/client.go
```diff
@@ -12,7 +12,20 @@ import (
 	"time"
 )
 
-var logger = log.ClientLogger
+var logger *log.Logger
```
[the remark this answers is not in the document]
miha-plesko: Also, what do you think about moving the variable from global scope to the Client struct (just like you did it for the Server)?

test/communication_test.go
```diff
@@ -25,15 +25,16 @@ var testGrpcClientConn *grpc.ClientConn
...
-	server, err := server.NewProtocolServer("testdata/server.pem", "testdata/server.key")
+	server, err := server.NewProtocolServer("testdata/server.pem", "testdata/server.key", "")
 	if err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
+
+	// Set debug log level for easier troubleshooting in case something goes wrong
+	client.SetLogLevel("notice")
+	server.SetLogLevel("notice")
```
miha-plesko: Here in the test you could then have yet another logger implemented that would just remember all the messages obtained so far (without printing them) and only print them if an error occurs 😄

log/logger.go
```diff
@@ -0,0 +1,68 @@
+package log
+
+import (
+	"github.com/op/go-logging"
+	"io"
+)
+
+// BaseLogger is a convenience interface that makes use of all functionality from go-logging's
+// Logger struct. In addition, it defines Disable() and SetLevel(level) functions
+type BaseLogger interface {
```
miha-plesko: May I suggest to name it simply `Logger`?

cli/server.go
```diff
@@ -26,11 +31,25 @@ var ServerCmd = cli.Command{
 }
 
 // startEmmyServer configures and starts the gRPC server at the desired port
-func startEmmyServer(port int, certPath, keyPath string) error {
-	srv, err := server.NewProtocolServer(certPath, keyPath)
+func startEmmyServer(port int, certPath, keyPath, logfilePath, loglevel string) error {
```
miha-plesko: `logfilePath, loglevel` -> `logFilePath, logLevel` (camelCase variable names)

cli/server.go
```diff
@@ -26,11 +31,25 @@ var ServerCmd = cli.Command{
...
 	if err != nil {
 		return err
 	}
+
+	srv, err := server.NewProtocolServer(certPath, keyPath, logger)
+	if err != nil {
+		return err
+	}
+
+	srv.SetLogLevel(loglevel)
```
miha-plesko: I'm not sure why we're polluting our `Server` with logging functions. I'd rather fully configure the logger and only then assign it to the `Server` struct. Specifically, I'd add following line just after L46:

```go
logger.SetLevel(logLevel)
```

server/server.go
```diff
@@ -106,15 +103,23 @@ func (s *Server) Teardown() {
 // in order to provide a nicer API when setting up the server.
 func (s *Server) EnableTracing() {
 	grpc.EnableTracing = true
-	logger.Notice("Enabled gRPC tracing")
+	s.logger.Notice("Enabled gRPC tracing")
+}
+
+func (s *Server) SetLogLevel(level string) error {
```
miha-plesko: I think we don't need this wrapper here, we are supposed to fully configure the logger prior even assigning it to the `Server`. Please remove.

server/server.go
```diff
@@ -106,15 +103,23 @@ func (s *Server) Teardown() {
...
+	return s.logger.SetLevel(level)
+}
+
+func (s *Server) DisableLogging() {
```
miha-plesko: Same here, please remove this function and disable logging on the logger prior assigning it to the struct instead.

client/client.go
```diff
@@ -12,7 +12,20 @@ import (
...
+// init instantiates and configures client logger with default log level.
+func init() {
+	clientLogger, err := log.NewStdoutLogger("client", log.INFO, log.FORMAT_SHORT)
+	if err != nil {
+		panic(err)
+	}
+	logger = *clientLogger
+}
+
+func DisableLogging() {
```
miha-plesko: Please remove this wrapper function `DisableLogging` and the `SetLogLevel` in L30. The logger should be configured upfront, so there is no need to bother our awesome `Client` with logging auxilary functions ;)

log/logger_file.go
```diff
@@ -0,0 +1,39 @@
...
+
+	leveledBackend := logging.SetBackend(backend)
+	leveledBackend.SetLevel(levelInt, module)
+	logger := &FileLogger{basicLogger}
+	logger.SetBackend(leveledBackend)
+
+	return logger, nil
+}
+
+func GetDefaultFileLogger(module, logfilePath string) (*FileLogger, error) {
```
miha-plesko: The `Default` in function name probably refers to default level and format being used? Even so, it's quite specialized already by `module` and `logFileLocation` (<-- camelCase 😈 ) so it might be better to shorten the name into:

```go
GetFileLogger
```

WDYT?

miha-plesko: EDIT: On a second thought, do we really need this function at all? I think whenever we use it, we then set logging level manually in the very next line with `SetLogLevel()`. I may be missing something, but wouldn't it be better to use the `NewFileLogger` in first place?

log/logger_stdout.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"github.com/op/go-logging"
+	"os"
+)
+
+// StdoutLogger outputs logs to standard output.
+type StdoutLogger struct {
+	basicLogger
+}
+
```
miha-plesko: You could add a line here that would ensure in compile time already that the logger implementation complies with your interface:

```go
var _ BaseLogger = (*StdoutLogger)(nil)
```

If you decide to do so, please do it for all other logger implementations as well.

log/logger_stdout_file.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	formattedStdoutBackend, err := basicLogger.setupFormattedBackend(os.Stdout, formatStdout)
+	if err != nil {
+		return nil, err
+	}
+	backends := []logging.Backend{formattedStdoutBackend}
+
+	// a non-empty string was provided as argument, which is assumed to be the absolute path
+	// to the file in which to write logs
```
miha-plesko: Is this comment still valid?

test/logging_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"testing"
+)
+
+func TestInvalidLogLevel(t *testing.T) {
+	_, err := log.NewStdoutLogger("test", "qwerty", log.FORMAT_SHORT)
+	assert.NotNil(t, err, "should produce an error due to invalid log level")
+}
+
+func TestInvalidFormat(t *testing.T) {
+	_, err := log.NewStdoutLogger("test", log.INFO, "blahblah")
```
miha-plesko: `"blahblah"` could be replaced with something more descriptive, say `"invalid-format"`

log/logger.go
mancabizjak: Sure, makes sense :slightly_smiling_face: . Done.

cli/server.go
mancabizjak: Oops. Fixed.

test/logging_test.go
mancabizjak: :D Sorry, got carried away. Okay, done.

log/logger_stdout_file.go
mancabizjak: Nope, it was left-over from previous implementation when we were checking whether `logFilePath != ""` Thanks.

cli/server.go
mancabizjak: I completely agree. Done.

server/server.go
mancabizjak: In addition, I am removing `func Disable()` on the `Logger` interface in favor of  newly introduced `struct NullLogger` that just discards all the logging messages.

log/logger_file.go
mancabizjak: Indeed. I went with your second comment and removed wrapper functions altogether.

client/client.go
mancabizjak: In addition, you will ntoice that L15 now reads `var Logger log.Logger` instead of `var logger log.StdoutLogger`, so that test functions can replace the client's default `StdoutLogger` with, for instance, newly introduced `NullLogger`. For this purpose, client package exports its logger (notice `Logger` instead of our previous `logger`).

log/logger.go
```diff
@@ -0,0 +1,68 @@
...
+	// These functions are our own
+	Disable()
```
mancabizjak: L12 is now removed in favor of `Logger`'s implementation `NullLogger`, whose sole function is to suppress logging output. This way loggers can just be replaced (for instance, when running test functions), instead of all having a `Disable()` function.

cli/server.go
```diff
@@ -26,11 +31,24 @@ var ServerCmd = cli.Command{
...
-func startEmmyServer(port int, certPath, keyPath string) error {
-	srv, err := server.NewProtocolServer(certPath, keyPath)
+func startEmmyServer(port int, certPath, keyPath, logFilePath, logLevel string) error {
+	var err error
+	var logger log.Logger
+
+	if logFilePath == "" {
+		logger, err = log.NewStdoutLogger("server", logLevel, log.FORMAT_LONG)
+	} else {
+		logger, err = log.NewStdoutLogger("server", logLevel, log.FORMAT_LONG_COLORLESS)
```
miha-plesko: I think `NewStdoutFileLogger` should be used here. At the moment `emmy server start --logfile ./burek.txt` does not log to the file given.

client/client.go
```diff
@@ -12,24 +12,29 @@ import (
 	"time"
 )
 
-var logger = log.ClientLogger
+var Logger log.Logger
```
miha-plesko: Is there any reason why we're now exporting the logger? Probably I'm just missing something.

docs/develop.md
```diff
@@ -4,6 +4,6 @@
 
 Go into the root project folder and execute:
 
-```
-protoc -I protobuf/ protobuf/msgs.proto --go_out=plugins=grpc:protobuf
+```bash
+$ protoc -I protobuf/ protobuf/messages.proto protobuf/services.proto protobuf/enums.proto --go_out=plugins=grpc:protobuf
```
miha-plesko: I'm not sure how modifying logging functionality could cause changes like this? 👼

protobuf/enums.proto
```diff
@@ -0,0 +1,29 @@
+syntax = "proto3";
```
miha-plesko: Um, what? 👼

test/qr_test.go
```diff
@@ -11,7 +11,8 @@ import (
 )
 
 func TestQRProof(t *testing.T) {
-	log.TurnOff() // otherwise too many communication messages
+	prevLogger := client.Logger
+	client.Logger = log.NewNullLogger()
```
miha-plesko: Now I see why you're exporting the logger. I was expecting a solution with implementing `func setLogger(newLogger)` on the client. I realize you probably did it this way since I was annoyed when seeing logger-related functions in client/client.go, but `setLogger` is different IMO. Anyway, do as you please, you can leave it with like this.

mancabizjak: Well yeah, this exactly, you basically answered yourself :smile: 
I have to admit I wasn't too fond of exporting the logger, so I'll implement your suggestion.

protobuf/enums.proto
mancabizjak: :confused:

cli/server.go
miha-plesko: Is it? 😄

log/logging.go
```diff
@@ -1,61 +1,69 @@
 package log
```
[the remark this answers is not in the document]
miha-plesko: Is it? 😄

mancabizjak:
@miha-plesko Last comments implemented, rebase done (successfully this time :wink:). Thanks a lot for a thorough review!

miha-plesko:
🎆 🍾 🎉

PR #29, opened 2017-10-04
Split protocol buffer definitions to multiple files

mancabizjak:
This PR splits `msgs.proto` to `enums.proto`, `messages.proto` and `services.proto`. `services.proto` imports `messages.proto`, while the latter imports `enums.proto`. Protocol buffer definitions are now clearly organized due to this logical separation.

Consequently,  the Go code generated by the `protoc` compiler  is also kept in three distinct `.pb.go` files.

The `protoc` command in `docs/develop.md` was also updated to reflect this change. 

This change does not affect any of the source files that import the `protobuf` package.


mancabizjak:
Yes @miha-stopar. Squashed, ready to merge :+1:

PR #30, opened 2017-10-10
Apply Apache v2.0 license

mancabizjak:
This commit adds LICENSE and AUTHORS files and adds the license header to all original source files.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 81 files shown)


mancabizjak:
@miha-plesko I think it's a matter of preference, it is certainly not required by the license (the only required stuff goes in LICENSE or NOTICE). But it is quite common. I was checking two other repos, go-grpc and one of go-grpc-ecosystem repos, which are both licensed under Apache v2.0 and both had the AUTHORS file.

But if the copyright notice in the comments of source files clearly states the author, then you should be okay. 

If @miha-stopar prefers putting the company name instead of "Emmy authors" in license header, I'm okay with it and will change :+1:

miha-stopar:
@mancabizjak I think XLAB is more appropriate.

mancabizjak:
Allright then, I'll remove AUTHORS and modify headers with the next commit.

mancabizjak:
Okay, it is done. I think we can merge @miha-stopar?

PR #31, opened 2017-10-16
Q-One-Way based commitments implemented.

miha-stopar:
Q-One-Way commitments based on RSA with bit commitment proof, preimage proof, and proof of multiplication property.


miha-stopar:
The same as rsa-based-commitment PR, however, due to the numerous problems when trying to rebase to incorporate the last changes from master (license,...), I opened a new PR because it was way faster than resolving conflicts.

PR #34, opened 2017-10-26
Make transcript a struct in dlogproofs package

### LINKED ISSUE #33: Transcript as struct
Currently, type Credential is implemented as struct with, among others, fields T1 and T2 as arrays of big.Int: ([https://github.com/xlab-si/emmy/blob/master/crypto/pseudonymsys/org_issue.go#L28](url))
```go
type Credential struct {
	SmallAToGamma *big.Int
	SmallBToGamma *big.Int
	AToGamma      *big.Int
	BToGamma      *big.Int
	T1            []*big.Int
	T2            []*big.Int
}
```
If I'm not wrong, the T1 and T2 arrays always hold the same structure (A, B, Hash, ZAlpha).
In my opinion, they should be represented as struct instead of big.Int arrays.

The obvious arguments are type safety and code readability.
Another reason (and the reason I came across this problem) is the binding to Java code, where arrays of strings are not supported (strings are used instead of big.Int for ease of conversion). Because of this, mobile library currently doesn't fully support the Credential type.

I propose to create a `type Transcript struct` holding A, B, Hash, ZAlpha values, and use this type in `Credential`.

mancabizjak:
This commit introduces Transcript struct in the dlogproofs package. It greatly improves readability. In addition, it will allow us to easily achieve compatibility with types supported by go tools for creating language bindings.

Resolves #33.

### DIFF
--- a/client/pseudonymsys.go
+++ b/client/pseudonymsys.go
@@ -248,16 +248,16 @@ func (c *PseudonymsysClient) TransferCredential(orgName string, userSecret *big.
 	x1, x2 := equalityProver.GetProofRandomData(userSecret, nym.A, credential.SmallAToGamma)
 
 	transcript1 := &pb.PseudonymsysTranscript{
-		A:      credential.T1[0].Bytes(),
-		B:      credential.T1[1].Bytes(),
-		Hash:   credential.T1[2].Bytes(),
-		ZAlpha: credential.T1[3].Bytes(),
+		A:      credential.T1.A.Bytes(),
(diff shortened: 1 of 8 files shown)


mancabizjak:
@zitnik Agreed. Patch for the EC coming soon.

miha-stopar:
Should be fine, just remove log.txt.

mancabizjak:
Oops :open_mouth: . Gone already :slightly_smiling_face:

miha-stopar:
Cool :)

PR #39, opened 2017-11-20
Expose functions for generation of master secret keys

mancabizjak:
Until now client's master secret key was loaded from the config file. This commit adds two simple functions, both named `GenerateMasterKey` to structs `PseudonymsysClient` and `PseudonymsysClientEC` from the `client` package. Client can now call these functions to generate the secret master key, needed for all the protocols of the schemes, on-the-fly. 

Fixed secret key and corresponding function for loading it were removed from the `config` package.


client/keygen.go
```diff
@@ -0,0 +1,20 @@
...
+// 0 and order of the group.
+func GenerateMasterKey(group *groups.SchnorrGroup) *big.Int {
+	return common.GetRandomInt(group.Q)
+}
+
+// GenerateMasterKeyEC generates a master secret key to use with elliptic curve schemes.
+func GenerateMasterKeyEC(curve dlog.Curve) *big.Int {
+	discreteLog := dlog.NewECDLog(curve)
+	return common.GetRandomInt(discreteLog.OrderOfSubgroup)
+}
```
miha-stopar: Not sure what would be the best place to put these two functions. The master keys will be generated differently for different schemes (where one scheme is pseudonymsys but for example new is coming that is based on RSA group), however at the same time they will probably always simply return a random element  < group.Order (as it is the case in these two functions). So we might simply use group.GetRandomElement. Note that each group should have such a method - Schnorr group has it, while EC group needs to be refactored (to expose the same methods as Schnorr group) and moved into groups package.

PR #40, opened 2017-11-20
TransferCredentials returns session key instead of bool

mancabizjak:
This commit adds support for returning a session key from pseudonymys's TransferCredential function, for both modular and EC implementation (up until now we were only returning `bool`s indicating whether authentication was successful or not). Server struct now embeds a new `sessionManager` struct tasked with generation of session keys. 

When session key cannot be returned (either because of failed authentication or some other error), `ProtocolError` field of the `Message` is set, while `SessionKey` (to be inserted within `Message.Content`) remains unset. Newly generated Go-code for protocol buffers is included.

You will notice that tests for the pseudonym system schemes are not ideal, and have grown quite a bit (not only as a result of this PR). I would prefer some refactoring to keep tests really unitary and therefore more manageable, but I'll handle this in another PR (remember, @miha-plesko? :wink:)


miha-plesko:
Oh, I remember pretty well @mancabizjak, my hand started shaking that day when a PR with **43** (that's fourty-three, yes) modified files stroke me out of blue ⚡️ 🌩 😄

PR #41, opened 2017-11-30
EC group moved into groups package.

miha-stopar:
EC group moved; functions changed to have inputs/outputs as types.ECGroupElement; functions renamed to be compliant with Schnorr group.


types/types.go
```diff
@@ -49,6 +49,14 @@ func ToPbECGroupElement(el *ECGroupElement) *pb.ECGroupElement {
 	return &x
 }
 
+func CmpECGroupElements(a, b *ECGroupElement) bool {
+	if a.X.Cmp(b.X) == 0 && a.Y.Cmp(b.Y) == 0 {
```
miha-stopar: I see the first comment coming :)

server/server.go
```diff
@@ -74,7 +74,7 @@ func NewProtocolServer(certFile, keyFile string, logger log.Logger) (*Server, er
 			grpc.MaxConcurrentStreams(math.MaxUint32),
 			grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
 		),
-		logger: logger,
+		logger:         logger,
```
zitnik: This is minor, but I think it would look nicer without the extra spaces, or, if you wish to have all arguments lined up, I believe lines 72-76 (starting with `grpc.`) should be indented as well.

types/types.go
zitnik: Well, you figured this one yourself :)

Why not simply `return a.X.Cmp(b.X) == 0 && a.Y.Cmp(b.Y) == 0` ?

server/server.go
miha-stopar: That was actually done by gofmt, so I would leave as it is, otherwise it will change it next time again.

miha-stopar:
Regarding interface - it will be removed for ZnGroup as well. At first I wanted to have one interface for all groups, but it doesn't make much sense as Zn groups deal with *big.Int inputs and outputs, while EC groups deal with ECGroupElements. Additionally, it seems groups will actually have different methods (although some subset of methods will be common, I will check the possibility of interfaces at some later point again).

PR #43, opened 2017-12-06
Damgard-Fujisaki commitment

miha-stopar:
First version of Damgard-Fujisaki commitment scheme. Corresponding proofs (knowledge how to open a commitment ... ) are coming in next PR + some further refactoring of groups package.


crypto/commitments/damgard-fujisaki.go
```diff
@@ -0,0 +1,130 @@
...
+type DamgardFujisakiCommitter struct {
+	QRSpecialRSA 		*groups.QRSpecialRSA
+	H               *big.Int
+	G               *big.Int // G = H^alpha % N
+	committedValue  *big.Int
+	r               *big.Int
+}
+
+func NewDamgardFujisakiCommitter(n, h, g *big.Int) *DamgardFujisakiCommitter {
+	group := groups.NewQRSpecialRSAPublic(n)
```
mancabizjak: You can remove L57 and put `groups.NewQRSpecialRSAPublic(n)` directly in L59.

crypto/commitments/damgard-fujisaki.go
```diff
@@ -0,0 +1,130 @@
...
+		QRSpecialRSA: group,
+		H: h,
+		G: g,
+	}
+}
+
+func (committer *DamgardFujisakiCommitter) GetCommitMsg(a *big.Int) (*big.Int, error) {
+	group := committer.QRSpecialRSA
+	if a.Cmp(group.N) != -1 {
+		err := errors.New("the committed value needs to be < N") // TODO: to be checked
```
mancabizjak: I don't understand the TODO. Will you implement it in scope of this PR?
EDIT: okay, according to our offline discussion, I bet this is indeed coming in one of the follow-up PRs.

crypto/common/primes.go
```diff
@@ -39,6 +39,26 @@ func GetSafePrime(bits int) (p *big.Int, err error) {
 	}
 }
 
+// GetSpecialRSAPrimes returns primes P, Q, p, q such that P = 2*p + 1 and Q = 2*q + 1.
+func GetSpecialRSAPrimes(bits int) (P, Q, p, q *big.Int, err error) {
```
mancabizjak: For better readability, we could maybe make a `struct SpecialRSAPrimes` with fields `P,Q,p,q`, and return this struct here. In my opinion, it's better to return two values from a function (pointer to a struct and error) instead of 4 `*big.Int`s and error :slightly_smiling_face: At least from perspective of the caller, who has to deal with these returned values (because he has to be aware of the order in which these values are returned...)

crypto/groups/qr_special_rsa.go
```diff
@@ -0,0 +1,123 @@
...
+	Order *big.Int // order of group QR_N (it is SmallP * SmallQ)
+}
+
+func NewQRSpecialRSA(safePrimeBitLength int) (*QRSpecialRSA, error) {
+	P, Q, p, q, err := common.GetSpecialRSAPrimes(safePrimeBitLength)
+	if err != nil {
+		return nil, err
+	}
+	N := new(big.Int).Mul(P, Q)
+	order := new(big.Int).Mul(p, q)
```
mancabizjak: I would move right-hand side of expressions on L44 and L45 to L47 and L62, respectively.

crypto/commitments/damgard-fujisaki.go
miha-stopar: It needs to be checked which values can be committed to (will do it in next PR).

crypto/commitments/damgard-fujisaki.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"github.com/xlab-si/emmy/crypto/common"
+	"math/big"
+	"errors"
+	"github.com/xlab-si/emmy/crypto/groups"
+)
+
+func DamgardFujisaki() (bool, error){
+	safePrimeBitLength := 1024
```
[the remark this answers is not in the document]
miha-stopar: There is a function in emmy at the top of the each file containing crypto primitive that demonstrates the usage of this primitive. DamgardFujisaki is such a function, so I think having it like it is makes it more simple to understand how to use it. In any case, it is just a demonstrator.

miha-stopar:
@mancabizjak @zitnik Thanks for a review, I think all comments have been applied.

crypto/commitments/damgard-fujisaki.go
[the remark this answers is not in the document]
zitnik: OK. I don't really like the demonstrator functions (here), but that's a problem for another PR/issue :)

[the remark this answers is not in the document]
miha-stopar: I know. There are many things I don't like in there, but it is a process. Almost every time I add a new crypto primitive I see things which should be changed or improved. Some of these things I do in parallel (like groups package right now), others will wait until all primitives for anonymous credentials are there, for example refactoring of proofs methods (at least GetProofRandomData naming), technique to convert sigma protocols into ZKPs will be changed and provided for all proofs, docs need to be updated ...

PR #45, opened 2017-12-20
Dlog retired

miha-stopar:
Groups package refactored. Preimage knowledge proofs refactored (renaming) and switch to binary challenges. Started working on documentation.


mancabizjak:
I'm good with all the changes. 

The only thing I want to comment on is the usage of `^` for denoting exponents in `docs/theory.md`. I remember we used subscript text like this *A<sub>b</sub>* before, maybe we could apply superscript notation for exponents in a similar way, for instance *g<sup>z</sup>*?

miha-stopar:
@mancabizjak @zitnik I addressed both comments and squashed the commits, I think it is ready to merge :). Thanks!

miha-stopar:
@zitnik I wouldn't do it if there were any significant changes, but yes, I won't do it again. Didn't know about "squash and merge".

PR #48, opened 2018-01-05
Retire types package

mancabizjak:
This PR retires emmy's `types` package and moves structs and function definitions previously kept in `types/types.go` to more appropriate places. Why? Well I've been looking into [Style guideline for Go packages](https://rakyll.org/style-packages/) where it is clearly explained
1. that types should be defined close to where they are used,
2. that in Go, types and functions should be organized by their functional responsibility (e.g. we should avoid generic packages like our `types` package).

So far our `types.go` was in a separate package and we put there:
* some common types used by crypto packages as well as communication code from client and server packages,
* functions for translating types from what I refer to as *native emmy types* to their protocol buffer alternatives (and vice versa).

And here's a summary of most notable changes that emerged from not having `types` package anymore:
* definition of `struct ECGroupElement` was moved to `groups` package (there we already had `struct ECGroup` defined, so it seemed appropriate)
* introduction of `translations.go` in the `protobuf` package for handling translations from/to protobuf and emmy types. Since protobuf types are defined in the same package, I hooked a function `GetNativeType()` to them. As for the other way around, I just exported global functions.
* `struct ServiceInfo` and its initializer were moved to `client/info.go`. Because of this, I had to modify the function `LoadServiceInfo` from `config/config.go` to not return `struct ServiceInfo` but three `strings` in order to avoid circular import.
* a new package `protocoltypes`  in `crypto/zkp` with definition of constants for sigma/ZKP/ZKPOK.
* a new file `basic_types.go` in `crypto/common` package with definition of `Pair` and `Triple` structs and their initializers. Now this is exactly what I wanted to avoid, but I don't see a better option.

I'm aware there's a lot of changes, but it's all just refactoring and moving the source to more appropriate places. Any suggestions on how to further improve the changes are very welcome :slightly_smiling_face:

PS: We still have to equip a lot of code with some comments. Perhaps I'll push a commit that does this, at least for the refactored/new functions.


server/pedersen_ec.go
```diff
@@ -51,7 +50,7 @@ func (s *Server) PedersenEC(curveType groups.ECurve, stream pb.Protocol_RunServe
 		return err
 	}
 
-	el := types.ToECGroupElement(ecgrop)
+	el := ecgrop.GetNativeType()
```
miha-stopar: Why not ecgroup? Or perhaps ecGroup?

mancabizjak: :smile:  I have no idea, it's been like that for a while. I changed the names to to `el` (pb type) and `myEl` (emmy type) so that its similar as with `h` and `myH` above.

crypto/groups/ec.go
```diff
@@ -34,6 +33,24 @@ const (
...
+}
+
+func NewECGroupElement(x, y *big.Int) *ECGroupElement {
+	return &ECGroupElement{
+		X: x,
+		Y: y,
+	}
+}
+
+//TODO Is this name appropriate? Perhaps 'Equals' would be better?
```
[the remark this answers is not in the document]
miha-stopar: Agree.

mancabizjak:
Thank you both! I did the renaming of `ECGroupElement`'s `Cmp` function to `Equals`, I think we're good now :+1:

PR #51, opened 2018-01-09
Make error statements and their formatting consistent

mancabizjak:
This commit re-formats error messages in `fmt.Errorf()` calls in order to conform to [code review recommendations for formatting error messages in Go](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

All the errors are now produced with `fmt.Errorf()` to achieve better consistency across the project. Calls to `errors.New()` were removed in favor of `fmt.Errorf()` calls.

I propose we follow the same convention (e.g. the use of `fmt.Errorf()` for definition of errors, and using short error messages starting with lower-cased letters and no punctuation at the end) in all new additions to the project. If you agree, we should "enforce" this during code review for all the subsequent PRs :slightly_smiling_face:

### DIFF
--- a/crypto/commitments/rsa_based.go
+++ b/crypto/commitments/rsa_based.go
@@ -19,7 +19,6 @@ package commitments
 
 import (
 	"crypto/rand"
-	"errors"
 	"fmt"
 	"math/big"
 
@@ -78,7 +77,7 @@ func NewRSABasedCommitter(homomorphism func(*big.Int) *big.Int, homomorphismInv 
 	// Note that for other q-one-way homomorphisms this validation would be different:
 	yIsValid := H.IsElementInGroup(y)
 	if !yIsValid {
-		return nil, fmt.Errorf("Y is not valid")
+		return nil, fmt.Errorf("y is not valid")
 	}
 	return &RSABasedCommitter{
 		Homomorphism:    homomorphism,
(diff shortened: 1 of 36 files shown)


crypto/commitments/rsa_based.go
```diff
@@ -43,7 +42,7 @@ func GenerateRSABasedQOneWay(nBitLength int) (func(*big.Int) *big.Int, func(*big
 		return nil, nil, nil, nil, err
 	}
 	if q.Cmp(rsa.N) < 1 {
-		return nil, nil, nil, nil, fmt.Errorf("Q must be > N")
+		return nil, nil, nil, nil, fmt.Errorf("q must be > n")
```
zitnik: I would rather have the names of mathematical variables in the same case as they are in comments. Which means upper-case `Q` and `N` here.

crypto/groups/qr_rsa.go
```diff
@@ -36,7 +37,7 @@ type QRRSA struct {
 
 func NewQRRSA(P, Q *big.Int) (*QRRSA, error) {
 	if !P.ProbablyPrime(20) || !Q.ProbablyPrime(20) {
-		return nil, errors.New("P and Q must be primes")
+		return nil, fmt.Errorf("p and q must be primes")
```
zitnik: Same here.

crypto/commitments/rsa_based.go
mancabizjak: Huh, you're right. Done!

crypto/groups/qr_rsa.go
mancabizjak: Done!

PR #52, opened 2018-01-10
Retire test package

mancabizjak:
This PR continues the work started in #49 and retires the `test` package. The remaining tests previously kept in the `test` package were reorganized and moved to more appropriate packages (`client` and `log`). Regarding communication tests (e.g. the ones that require gRPC server and clients): we have a common server setup and teardown code, along with connection establishment in the `client` package's `TestMain` function. As for testing individual protocols/schemes, I put them in the same package but more appropriate `_test` files.

Apart from the changes that were required as a consequence of code reorganization, I didn't apply any changes to the test code itself. 

I'll be happy to hear your doubts/suggestions for further improvement.


mancabizjak:
I provided a relatively small and nasty workaround in order to make tests pass (but it still involved 4 files getting changed in the process...). 

The CSPaillier tests were always problematic, and in future the crypto code for the protocol + unit tests and communication layer should be refactored; I started doing it myself but I believe this is an extremely low priority now, so I gave it up.

PR #53, opened 2018-01-12
Damgard-Fujisaki associated ZKPs

miha-stopar:
DF scheme fixed. Two associated ZKPs implemented:
- Proof that you know how to open a commitment: given a commitment c, prove that you know x, r such that c = g^x * h^r (mod N).
- Proof that x3 = x1*x2 given three commitments c1, c2, c3 where:
  c1 = g^x1 * h^r1 (mod N)
  c2 = g^x2 * h^r2 (mod N)
  c3 = g^x3 * h^r3 (mod N)

### DIFF
--- a/crypto/zkp/primitives/commitments/damgard-fujisaki_test.go
+++ b/crypto/zkp/primitives/commitments/damgard-fujisaki_test.go
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2017 XLAB d.o.o.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package commitmentzkp
+
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/xlab-si/emmy/crypto/commitments"
+	"github.com/xlab-si/emmy/crypto/common"
+	"math/big"
+	"testing"
+)
+
+// TestProveDFCommitmentOpening demonstrates how to prove that you can open DamgardFujisaki commitment.
+func TestProveDFCommitmentOpening(t *testing.T) {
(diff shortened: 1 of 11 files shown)


crypto/zkp/primitives/commitments/damgard-fujisaki_test.go
```diff
@@ -0,0 +1,117 @@
...
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/xlab-si/emmy/crypto/commitments"
+	"github.com/xlab-si/emmy/crypto/common"
+	"math/big"
+	"testing"
+)
+
+// TestProveDFCommitmentOpening demonstrates how to prove that you can open DamgardFujisaki commitment.
+func TestProveDFCommitmentOpening(t *testing.T) {
```
mancabizjak: I suggest renaming this test to something that includes `DamgardFujisaki`. For instance, `TestOpenDamgardFujisakiCommitment` (or whatever you find most appropriate). It produces a clearer output when the tests are actually run.

crypto/zkp/primitives/commitments/damgard-fujisaki_test.go
```diff
@@ -0,0 +1,117 @@
...
+	challenge := verifier.GetChallenge()
+	s1, s2 := prover.GetProofData(challenge)
+	proved := verifier.Verify(s1, s2)
+
+	assert.Equal(t, true, proved, "DamgardFujisaki opening proof failed.")
+}
+
+// TestProveDFCommitmentMultiplication demonstrates how to prove that for given commitments
+// c1 = g^x1 * h^r1, c2 = g^x2 * h^r2, c3 = g^x3 * h^r3, it holds x3 = x1 * x2
+func TestProveDFCommitmentMultiplication(t *testing.T) {
```
mancabizjak: I suggest renaming this test to something that includes `DamgardFujisaki`. For instance, `TestProveDamgardFujisakiCommitmentMultiplication`. It is indeed longer but produces a clearer output when the tests are actually run.

miha-stopar:
I think I addressed all comments. I haven't added DamgardFujisaki New* function because it is only to be used internally when initializing DamgardFujisakiCommitter and DamgardFujisakiReceiver, instead I made it internal (now damgardFujisaki).

mancabizjak:
Thanks, great. 

I have two more comments (sorry, I didn't notice the first one before)
* please create test files `damgard-fujisaki_multiplication_test.go` and `damgard-fujisaki_opening.go` and put each unit test to appropriate file. And remove the file `damgard-fujisaki_test.go` that contains both of the tests. 
* Please also update comments above tese two tests, so that correct names are used (the old ones are still present).

PR #54, opened 2018-01-15
Support for less generic services and RPCs 

mancabizjak:
This is a rather large PR, whose purpose is to allow for a better definition of the protobuf services and their RPCs . Up until now, we had all the cryptographic schemes and protocols implemented with a single service and RPC:
```go
// A generic service
service Protocol {
	rpc Run (stream Message) returns (stream Message) {}
}
```
and to invoke a specific sigma protocol, we always had to pass a `SchemaType` in the first message to the server. The server then checked for this value and called appropriate functions to handle the protocol execution. I feel that this not the right approach anymore, as it seems too generalized.

Commits in this PR introduce the required changes to support more specific and intuitive (less generic) services and RPCs. For instance, we now have two additional services:
```go
service PseudonymSystemCA {
	rpc GenerateCertificate(stream Message) returns (stream Message) {}
	rpc GenerateCertificate_EC(stream Message) returns (stream Message) {}
}

service PseudonymSystem {
	rpc GenerateNym (stream Message) returns (stream Message) {}
	rpc GenerateNym_EC (stream Message) returns (stream Message) {}
	rpc ObtainCredential (stream Message) returns (stream Message) {}
	rpc ObtainCredential_EC (stream Message) returns (stream Message) {}
        ...
}
```
To support these changes and still keep the generic communication code (concerning especially client-side), quite a few changes had to be implemented. I recommend that you review these changes commit-by-commit; I tried to be descriptive in the commit messages.

To highlight the most important changes that emerged as a consequence of the above:
* The entire pseudonym system and pseudonym system CA schemes/protocols (both modular and EC) were ported to run over RPCs within newly defined protobuf services.
* Simple sigma protocols are still run over the generic service, but will be ported with an additional commit to this PR, as soon as we converge.
* **server-side**: in addition to having a `Run` handler, we now have pseudonym system specific handlers (server-side implementations of RPCs). You will also notice `EmmyServer` interface that basically aggregates all the interfaces implemented by our `Server`.
* **client-side**: here are the most significant changes. Since we no longer have only one client stream type (`Protocol_RunClient`), but several of them (`Protocol_RunClient`, `PseudonymSystemCA_GenerateCertificateClient`, `PseudonymSystem_TransferCredentialClient`, ...), communication code, especially the `func (c *genericClient) openStream()` had to be modified. `openStream` now accepts a `grpcClient` and a `string` which denotes the name of the method to call on the `grpcClient` in order to retrieve the client stream. The `grpcClient` field was moved from `struct genericClient` to each individual protocol client, since for the same reason we now need to support not only different client streams, but different grpc clients as well.


client/pseudonymsys_ca.go
```diff
@@ -30,28 +30,29 @@ import (
...
 
 	return &PseudonymsysCAClient{
-		genericClient: *genericClient,
+		genericClient: newGenericClient(),
+		grpcClient:    pb.NewPseudonymSystemCAClient(conn),
 		prover:        dlogproofs.NewSchnorrProver(group, protocoltypes.Sigma),
 	}, nil
 }
 
 // ObtainCertificate provides a certificate from trusted CA to the user. Note that CA
```
zitnik: Correct comment to `GenerateCertificate`.
On the other hand, I'm not sure I like the new name. How come you decided to rename this method?

client/pseudonymsys_ca_ec.go
```diff
@@ -30,32 +30,31 @@ import (
...
 
 	return &PseudonymsysCAClientEC{
-		genericClient: *genericClient,
+		genericClient: newGenericClient(),
+		grpcClient:    pb.NewPseudonymSystemCAClient(conn),
 		prover:        prover,
 	}, nil
 }
 
 // ObtainCertificate provides a certificate from trusted CA to the user. Note that CA
```
zitnik: Same here.

client/client.go
```diff
@@ -23,10 +23,13 @@ import (
 	"math/rand"
 	"time"
 
+	"golang.org/x/net/context"
```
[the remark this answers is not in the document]
mancabizjak: Thanks, I wasn't aware of it :+1: Done, I modified this import path in `client/client.go`, `client/info.go`, `server/info.go`

client/client.go
```diff
@@ -140,27 +139,56 @@ func (c *genericClient) getResponseTo(msg *pb.Message) (*pb.Message, error) {
...
+// grpcClients (each generated from its own RPC service), each has its own streamGenFunc(s)
+// (generated from the appropriate RPC within the service), it is the caller's responsibility
+// to provide appropriate grpcClient and streamGenFunc.
 // This function has to be called explicitly at the beginning of the protocol execution function.
-func (c *genericClient) openStream() error {
-	stream, err := c.protocolClient.Run(context.Background())
+func (c *genericClient) openStream(grpcClient interface{}, streamGenFunc string) error {
+	// Create structs compatible with reflect package
+	client := reflect.ValueOf(grpcClient)                            // we want to call streamGenFunc on this struct
+	params := []reflect.Value{reflect.ValueOf(context.Background())} // we want to pass these params to streamGenFunc
```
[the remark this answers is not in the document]
mancabizjak: `params` indeed represents a set of arguments to be passed to `streamGenFunc` (`streamGenFunc` is actually a general representation of RPCs like `Run`, `GenerateNym`, `GenerateNym_EC`,...). 

However, `params` is not an empty set of arguments; on L152, you will notice that we're actually using reflection to construct a slice of `reflect.Value`s that has a single element, which is `reflect.Value` representation of `context.Background()`. The result of `context.Background()` call is a value that we pass to all of the above mentioned `streamGenFunc`s  (see L161)  - we always use the same value regardless of the actual `streamGenFunc` being called.

So, `openStream` doesn't have a `params` argument because we don't expect it to be different between invocations of different `streamGenFunc`s - all the `streamGenFunc`s expect a `context.Context` argument, which can indeed be something other than the result of `context.Background()` call, but at this point, we never used anything else, so the same value is used during invocation of the `streamGenFunc`s.

client/pseudonymsys_ca.go
mancabizjak: Done, corrected comment here and in EC implementation as well.

As for the name, I actually modified server-side handlers (that were previously named `PseudonymsysCA` and `PseudonymsysCAEC`) before I got to the clients. So I noticed afterwards that on client-side they were indeed named `ObtainCertificate`. But I think the distinction is subtle, and since we have `GenerateNym`, and not `ObtainNym`, why not have `GenerateCertificate` as well? So I just went for consistency, although I have to admit I didn't put too much thought in it :slightly_smiling_face:

client/pseudonymsys_ca_ec.go
mancabizjak: Done.

protobuf/services.proto
```diff
@@ -26,6 +26,20 @@ service Protocol {
...
+	rpc GenerateNym_EC (stream Message) returns (stream Message) {}
+	rpc ObtainCredential (stream Message) returns (stream Message) {}
+	rpc ObtainCredential_EC (stream Message) returns (stream Message) {}
+	rpc TransferCredential (stream Message) returns (stream Message) {}
+	rpc TransferCredential_EC (stream Message) returns (stream Message) {}
+}
+
 service Info {
 	rpc GetServiceInfo(EmptyMsg) returns (ServiceInfo) {}
 }
```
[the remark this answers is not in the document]
mancabizjak: :open_mouth: Fixed. My new IDE only inserts them for Go files, not all the text files...

client/pseudonymsys_ca.go
zitnik: Well, we don't have total consistency anyway. There's `GenerateNym`, `ObtainCredentials`, and `Obtain/GenerateCertificate`. It would seem logical to me that "obtain" implies getting something from the server, while "generate" means a local operation - although I know it's not like this.
And since names `GenerateNym` and `ObtainCredentials` are used in the protocol definitions, we should keep them. I guess it doesn't really matter for Certificates then :)

mancabizjak:
@zitnik I had to revert one of the corrected imports of `context` back to `golang.org/x/net/context` (specifically, the import in `server/info.go`,  because it contains the `GetServiceInfo` handler, which is the only RPC handler that actually needs to be passed a `context.Context` struct and not a stream). 

The error ocurred because in the file `protobuf/services.pb.go`, which is auto-generated by protoc compiler, we have an import of `golang.org/x/net/context` and our recently renamed import causes an inconsistency:

```bash
$ go install
# github.com/xlab-si/emmy/server
server/server.go:54:5: cannot use (*Server)(nil) (type *Server) as type EmmyServer in assignment:
        *Server does not implement EmmyServer (wrong type for GetServiceInfo method)
                have GetServiceInfo("context".Context, *protobuf.EmptyMsg) (*protobuf.ServiceInfo, error)
                want GetServiceInfo("golang.org/x/net/context".Context, *protobuf.EmptyMsg) (*protobuf.ServiceInfo, error)
server/server.go:157:25: cannot use s (type *Server) as type protobuf.InfoServer in argument to protobuf.RegisterInfoServer:
        *Server does not implement protobuf.InfoServer (wrong type for GetServiceInfo method)
                have GetServiceInfo("context".Context, *protobuf.EmptyMsg) (*protobuf.ServiceInfo, error)
                want GetServiceInfo("golang.org/x/net/context".Context, *protobuf.EmptyMsg) (*protobuf.ServiceInfo, error)
```

I am not sure how come this was not caught in the previous tests.

PR #55, opened 2018-01-17
Make client connections more flexible

mancabizjak:
This PR introduces a `struct ConnectionConfig` that holds configuration information required for client to contact the server: _endpoint_, _CA certificate_ and TLS _insecurity_ parameter are now passed to client's
`GetConnection` function within a single struct. 

Moreover, the more important part of this PR allows for more flexibility with respect to the CA certificate, which should now be passed as a `[]byte` instead of having to pass the path to certificate on the client's filesystem. Some additional tests were also added in `client/tls_test.go`.


mancabizjak:
Thanks so much for this @zitnik. I too struggled with `InsecureSkipVerify` when I first implemented the code in question. I believe what you suggest makes total sense, and I'm okay with making these changes within this PR, since it will unavoidably require changes on the newly added code.

I'm drafting a better approach and will push a couple of commits soon.

mancabizjak:
@zitnik please take a look at the updated version. Looking forward to hearing your opinion :slightly_smiling_face:

PR #57, opened 2018-01-19
Improve server-side handling of protocol errors

mancabizjak:
This PR improves the way protocol errors are handled by emmy server.

So far, we only reported `ProtocolError`s back to the client in case any of pseudonym system protocols failed. However, these errors were not consistently logged at the server side. Here's a summary of changes:
* added logging for all `ProtocolError`s at the server side with the *error* severity level - so far we were only able to see these errors in one of the protocols, and only if the log level of `Server`'s logger was *debug*. Now we'll be able to detect these errors more quickly, as they will be shown in the server's log even if it's not running in debugging mode.
* the `Server`'s `send()` function is updated to more precisely log the type of message that was sent (either `msg.Content` or `msg.ProtocolError`). 
* some unnecessary payload initialization (passing something like `Content: nil` or even `Content: &pb.SomeType{}` to initialization of `pb.Message`) in case of protocol error ocurred was also removed.

### DIFF
--- a/server/pseudonymsys.go
+++ b/server/pseudonymsys.go
@@ -46,7 +46,7 @@ func (s *Server) PseudonymsysGenerateNym(req *pb.Message, stream pb.Protocol_Run
 	var resp *pb.Message
 
 	if !regKeyOk || err != nil {
-		s.logger.Errorf("registration key %s ok=%t, error=%v",
+		s.logger.Debugf("registration key %s ok=%t, error=%v",
 			proofRandData.RegKey, regKeyOk, err)
 		resp = &pb.Message{
 			ProtocolError: "registration key verification failed",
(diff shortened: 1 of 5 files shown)


mancabizjak:
I was thinking and I now see your reservations about this. 

Perhaps lowering the logging level for reporting protocol errors from *error* to *debug* makes more sense? I mean, as you said, failed registration attempts along with corresponding registration keys normally wouldn't (and shouldn't) be logged in the server's console as errors.

So, since the purpose of having any kind of logging of such protocol errors is to help identify the problems within development, do you feel *debug* logging level is more appropriate? I mean, in any kind of production setting you wouldn't run the server in the debug mode anyway, so the messages won't be visible by default.

PR #61, opened 2018-01-30
Add Dockerfile and docker-compose file

mancabizjak:
This commit adds a Dockerfile for building emmy server image. In addition, docker-compose file is added that starts emmy server container as well as redis container. 

These are added solely for convenience during development; the use of dockerized components is not assumed during unit testing on a given host or for Travis CI builds.

Brief guidelines and instructions were added to `docs/develop.md`.

### DIFF
--- a/docs/develop.md
+++ b/docs/develop.md
@@ -19,7 +19,18 @@ spin up an instance of emmy server.
 In addition, we provide a *docker-compose.yml* file you can use to start both emmy server as well as
 a redis database to hold registration keys. Note that registration keys need to exist in redis 
 beforehand; emmy server only checks for their existence, while a separate entity needs to insert
-them into redis (either you put them there manually, or some third party application has to).
+them into redis (either you put them there manually, or some third party application has to). The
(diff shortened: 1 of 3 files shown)


docs/develop.md
```diff
@@ -12,6 +12,27 @@ Go into the root project folder and execute:
...
 ```
 
+# Using dockerized Emmy server and redis for development
+For testing and ease of development this repository comes with a *Dockerfile* that you can use to 
+spin up an instance of emmy server. 
+
+In addition, we provide a *docker-compose.yml* file you can use to start both emmy server as well as
+a redis database to hold registration keys. Note that registration keys need to exist in redis 
+beforehand; emmy server only checks for their existence, while a separate entity needs to insert
+them into redis (either you put them there manually, or some third party application has to).
```
zitnik: If it's simple/short enough, can you add a command for inserting a registration key into redis?

mancabizjak: Done!

PR #64, opened 2018-02-07
Update loading of QRRSA from config

mancabizjak:
This PR cleans up unused configuration and updates function for loading QRRSA configuration. 

Previously we had `qr` and `qrsmall` in the configuration YAML, but we never used `qr`. I replaced the values of the previous `qr` with those defined for `qrsmall`, so that they are now always loaded (without the need for the caller to specify which one). `qrsmall` field was removed, as its values are now under `qr`.


mancabizjak:
@miha-stopar Travis CI build times out before all the tests can be completed for QR/QNR proofs with longer keys. I would keep the shorter ones, as the entire test suite already runs quite some time. Perhaps we should add safety controls to ensure a safe key size some time in the future. Do you agree?

miha-stopar:
OK. Perhaps just add a comment in the test that says that at least 1024 bit length should be used for p and q outside the tests.

PR #68, opened 2018-02-21
Use Makefile for a simplified build

mancabizjak:
This PR adds a *Makefile* with several targets for ease of testing and development. Instructions in the Travis CI file were updated so that `make` commands are used. 

Most importantly, documentation in `docs/develop` was also updated and improved on several places, so that the possibility of using `make` targets is reflected where appropriate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,62 @@
+.PHONY: setup setup_test setup_mobile setup_linter install test fmt lint android proto clean run
+
+ALL=./...
+
+all: install
+
+# Setup and update all the required tools
+setup: setup_test setup_linter setup_mobile
+
+setup_test:
+	go get -u github.com/stretchr/testify/assert
+
+setup_mobile:
+	go get -u golang.org/x/mobile/cmd/gomobile
+
+setup_linter:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,62 @@
...
+all: install
+
+# Setup and update all the required tools
+setup: setup_test setup_linter setup_mobile
+
+setup_test:
+	go get -u github.com/stretchr/testify/assert
+
+setup_mobile:
+	go get -u golang.org/x/mobile/cmd/gomobile
```
zitnik: I believe that `gomobile init` instruction is needed after getting gomobile here. Before calling init, gomobile bind will not work.

mancabizjak: Yes indeed. I had this in mind actually. Previously I had a single `android` target that had both `gomobile init` and `gomobile bind` commands, but I removed the `gomobile init` one because it really only has to be called once (and it takes quite a while to execute). I later also split the `setup` target into several sub-targets, including `setup_mobile`, and it makes perfect sense to have `gomobile init` there, I missed it.

Done - thanks!

PR #73, opened 2018-03-02
DamgardFujisaki commitment equality and square proof

miha-stopar:
DFCommitmentEqualityProof proves that the two commitments c1, c2 hide the same number. Given c1, c2, prove that c1 = g1^x * h1^r1 (mod n1) and c2 = g2^x * h2^r2 (mod n2) for some x, r1, r2 (note that both commitments hide x).

DFCommitmentSquareProof proves that the commitment hides the square. Given c, prove that c = g^(x^2) * h^r (mod n).

### DIFF
--- a/crypto/commitments/damgard-fujisaki.go
+++ b/crypto/commitments/damgard-fujisaki.go
@@ -71,6 +71,7 @@ func NewDamgardFujisakiCommitter(n, h, g, t *big.Int, k int) *DamgardFujisakiCom
 		T: t}
 }
 
+// TODO: the naming is not OK because it also sets committer.committedValue and committer.r
 func (committer *DamgardFujisakiCommitter) GetCommitMsg(a *big.Int) (*big.Int, error) {
 	abs := new(big.Int).Abs(a)
 	if abs.Cmp(committer.T) != -1 {
@@ -83,6 +84,18 @@ func (committer *DamgardFujisakiCommitter) GetCommitMsg(a *big.Int) (*big.Int, e
 	r := common.GetRandomInt(boundary)
 	c := committer.ComputeCommit(a, r)
 
+	committer.committedValue = a
+	committer.r = r
+	return c, nil
+}
+
+func (committer *DamgardFujisakiCommitter) GetCommitMsgWithGivenR(a, r *big.Int) (*big.Int, error) {
+	abs := new(big.Int).Abs(a)
+	if abs.Cmp(committer.T) != -1 {
+		return nil, fmt.Errorf("committed value needs to be in (-T, T)")
+	}
+	// c = G^a * H^r % group.N
+	c := committer.ComputeCommit(a, r)
 	committer.committedValue = a
 	committer.r = r
 	return c, nil
--- a/crypto/commitments/damgard-fujisaki.go
+++ b/crypto/commitments/damgard-fujisaki.go
@@ -143,17 +143,17 @@ func NewDamgardFujisakiReceiver(safePrimeBitLength, k int) (*DamgardFujisakiRece
 		nil
 }
 
-// NewDamgardFujisakiReceiverFromExisting returns an instance of receiver with the same
-// parameters as the receiver used as an input. Different instances are needed because
+// NewDamgardFujisakiReceiverFromParams returns an instance of a receiver with the
+// parameters as given by input. Different instances are needed because
 // each sets its own Commitment value.
-func NewDamgardFujisakiReceiverFromParams(qrSpecialRSA *groups.QRSpecialRSA, H, G *big.Int,
-	K int) (
+func NewDamgardFujisakiReceiverFromParams(qrSpecialRSA *groups.QRSpecialRSA, h, g *big.Int,
+	k int) (
 	*DamgardFujisakiReceiver, error) {
 	return &DamgardFujisakiReceiver{damgardFujisaki: damgardFujisaki{
 		QRSpecialRSA: qrSpecialRSA,
-		H:            H,
-		G:            G,
-		K:            K},
+		H:            h,
+		G:            g,
+		K:            k},
 	}, nil
 }
 
--- a/crypto/commitments/damgard-fujisaki.go
+++ b/crypto/commitments/damgard-fujisaki.go
@@ -133,13 +133,14 @@ func NewDamgardFujisakiReceiver(safePrimeBitLength, k int) (*DamgardFujisakiRece
 // NewDamgardFujisakiReceiverFromExisting returns an instance of receiver with the same
 // parameters as the receiver used as an input. Different instances are needed because
 // each sets its own Commitment value.
-func NewDamgardFujisakiReceiverFromExisting(receiver *DamgardFujisakiReceiver) (
+func NewDamgardFujisakiReceiverFromParams(qrSpecialRSA *groups.QRSpecialRSA, H, G *big.Int,
+	K int) (
 	*DamgardFujisakiReceiver, error) {
 	return &DamgardFujisakiReceiver{damgardFujisaki: damgardFujisaki{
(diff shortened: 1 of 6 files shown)


crypto/commitments/damgard-fujisaki.go
```diff
@@ -133,13 +133,14 @@ func NewDamgardFujisakiReceiver(safePrimeBitLength, k int) (*DamgardFujisakiRece
 // NewDamgardFujisakiReceiverFromExisting returns an instance of receiver with the same
 // parameters as the receiver used as an input. Different instances are needed because
 // each sets its own Commitment value.
-func NewDamgardFujisakiReceiverFromExisting(receiver *DamgardFujisakiReceiver) (
+func NewDamgardFujisakiReceiverFromParams(qrSpecialRSA *groups.QRSpecialRSA, H, G *big.Int,
```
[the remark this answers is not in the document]
mancabizjak: May I ask why this change was introduced? I mean, why are we now specifying every parameter instead of just copying the receiver's values? Test code is much more verbose now.

crypto/zkp/primitives/commitments/damgard-fujisaki_square.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func NewDFCommitmentSquareProver(committer *commitments.DamgardFujisakiCommitter,
+	x *big.Int, challengeSpaceSize int) (*DFCommitmentSquareProver, *big.Int, error) {
+
+	// Input committer contains c = g^(x^2) * h^r (mod n).
+	// We now create two committers - committer1 will contain c1 = g^x * h^r1 (mod n),
+	// committer2 will contain the same c as committer, but using a different base c = c1^x * h^r2.
+	// Note that c = c1^x * h^r2 = g^(x^2) * h^(r1*x) * h^r2, so we choose r2 = r - r1*x.
+	// DFCommitmentSquareProver proves that committer1 and committer2 hide the same value (x) -
```
mancabizjak: Can you move these comments to more appropriate place? For instance I think L39-L40 should be put above  `DFCommitmentSquareProver`.

crypto/zkp/primitives/commitments/damgard-fujisaki_square.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	prover := NewDFCommitmentEqualityProver(committer1, committer2, challengeSpaceSize)
+
+	return &DFCommitmentSquareProver{
+		equalityProver: prover,
+	}, c1, nil
+}
+
+func (prover *DFCommitmentSquareProver) GetProofRandomData() (*big.Int, *big.Int) {
```
mancabizjak: If you embed `*DFCommitmentEqualityProver` to `DFCommitmentSquareProver`, you can safely delete L70-L78, because in this case `DFCommitmentSquareProver` will not need to wrap `DFCommitmentEqualityProver`'s `GetProofRandomData()` and `GetProofData()` functions.

crypto/zkp/primitives/commitments/damgard-fujisaki_square.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"math/big"
+
+	"fmt"
+
+	"github.com/xlab-si/emmy/crypto/commitments"
+)
+
+type DFCommitmentSquareProver struct {
+	equalityProver *DFCommitmentEqualityProver
```
mancabizjak: Maybe it would be better to embed `*DFCommitmentEqualityProver` (e.g. just remove `equalityProver`).

crypto/zkp/primitives/commitments/damgard-fujisaki_square.go
```diff
@@ -0,0 +1,118 @@
...
+	return prover.equalityProver.GetProofRandomData()
+}
+
+func (prover *DFCommitmentSquareProver) GetProofData(challenge *big.Int) (*big.Int,
+	*big.Int, *big.Int) {
+	return prover.equalityProver.GetProofData(challenge)
+}
+
+type DFCommitmentSquareVerifier struct {
+	equalityVerifier *DFCommitmentEqualityVerifier
```
mancabizjak: Same as for `DFCommitmentSquareProver` - I suggest embedding `*DFCommitmentEqualityVerifier`.

crypto/zkp/primitives/commitments/damgard-fujisaki_square.go
```diff
@@ -0,0 +1,118 @@
...
+	receiver2.SetCommitment(receiver.Commitment)
+
+	verifier := NewDFCommitmentEqualityVerifier(receiver1, receiver2, challengeSpaceSize)
+
+	return &DFCommitmentSquareVerifier{
+		equalityVerifier: verifier,
+	}, nil
+}
+
+func (verifier *DFCommitmentSquareVerifier) SetProofRandomData(proofRandomData1,
```
mancabizjak: Please consider embedding `*DFCommitmentEqualityVerifier`, because it will allow you to avoid `L107-L118` - you will be able to safely erase them, while `DFCommitmentSquareVerifier` will be still able to use this functionality.

crypto/zkp/primitives/commitments/damgard-fujisaki_square_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/xlab-si/emmy/crypto/commitments"
+	"github.com/xlab-si/emmy/crypto/common"
+)
+
+// TestProveDamgardFujisakiCommitmentSquare demonstrates how to prove that the commitment
+// hides the square. Given c, prove that c = g^(x^2) * h^r (mod n).
+func TestProveDamgardFujisakiCommitmentSquare(t *testing.T) {
+	receiver, err := commitments.NewDamgardFujisakiReceiver(1024, 80)
```
mancabizjak: I know this probably won't be possible, but still - there's quite a lot pre-initialization logic for configuring receiver and committer, which are subsequently be passed as arguments to `NewDFCommitmentSquareProver` and `NewDFCommitmentSquareVerifier`; is it in any way possible (or does it make sense at all) to initialize committer/receiver within these functions?

The same comment goes for `NewDFCommitmentEqualityProver` and `NewDFCommitmentEqualityVerifier`.

crypto/commitments/damgard-fujisaki.go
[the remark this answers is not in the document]
miha-stopar: Changed to (h, g, k).

Because of this call:
receiver2, err := commitments.NewDamgardFujisakiReceiverFromParams(receiver.QRSpecialRSA, receiver.H, c1, receiver.K)

Note that there is c1 used instead of receiver.G. But also - receiver might not always be available.

crypto/zkp/primitives/commitments/damgard-fujisaki_square.go
miha-stopar: The comment describes the internals of the proof so the place is appropriate (it is not something that the user of a function needs to know, but I prefer to have such explanations available in the code). L39 and L40 depend on the lines before. But I added a comment to the prover struct.

crypto/zkp/primitives/commitments/damgard-fujisaki_square.go
miha-stopar: Good point!

crypto/zkp/primitives/commitments/damgard-fujisaki_square_test.go
miha-stopar: It somehow depends on the way this proofs will be used. Let's wait with this until I implement the new scheme - I might then change this in the way you proposed (if possible).

PR #74, opened 2018-03-02
Use dep for dependency management

janhartman:
This commit introduces the use of golang's tool for dependency management called dep to enable strict version control for emmy's dependencies. Dep uses the Gopkg.toml file to manage dependencies, which are stored in the vendor directory.

Some changes to Makefile, Dockerfile and .travis.yml were needed to make everything work with the vendor directory. The changes to the workflow when developing Emmy are described in docs/develop.md.

To summarize: the most frequent use case for dep is running `dep ensure`. The commands scans the imports in the source code and the Gopkg.toml file, then fetches or removes dependencies if necessary. 

As for the version restrictions: I constrained some of the dependencies to minor update range (i.e. 
we only allow semantic version patches) and some to master branch (these have not yet released critical changes which are present in master).
A short read of the [documentation](https://golang.github.io/dep/docs/daily-dep.html) is recommended.


Makefile
```diff
@@ -18,6 +24,11 @@ setup_linter:
 	go get -u github.com/alecthomas/gometalinter
 	gometalinter --install --update
 
+# Runs dep ensure to populate the vendor directory with
+# the required dependencies and potentially modify Gopkg.lock.
```
mancabizjak: modifies? :nerd_face:

Makefile
```diff
@@ -28,9 +39,9 @@ test:
 
 # Lists and formats all go source files with goimports
 fmt:
-	# List of files with different formatting than goimport's
-	goimports -l .
-	goimports -w .
+	# List of files with different formatting than goimports'
+	@goimports -l $(ALL_GO)
```
mancabizjak: Ah, thanks for bearing this mind, appreciated :+1:

docs/develop.md
```diff
@@ -42,6 +45,14 @@ Please install *goimports* and configure your source code editor to automaticall
...
  formatting of all go source files in the repository.  
 
+## Dependency management
+Emmy uses [dep](https://github.com/golang/dep) for dependency management. All direct dependencies 
+and their versions are specified in `Gopkg.toml`. At the moment, minor range locking is used (i.e. 
+we only allow semantic version patches. A quick read of `dep`'s [documentation](https://golang.github.io/dep/docs/daily-dep.html) 
+is recommended to become familiar with the tool. Note that for now, not all direct 
+dependencies are strictly version-locked (some of them have yet to create a release with necessary 
+features that are already implemented in the master branch).
+
```
mancabizjak: Since this document is a guide for developing & contributing to emmy, would you mind also adding some indication that `Gopkg.toml` and `Gopkg.lock` should both be versioned with the rest of the repository? 

I think we should make it clear for all the developers that if any vendoring changes are introduced, it is the responsibility of the developer who is making these changes to update `dep`'s files (although this might be obvious to the folks regularly using `dep`, this is a reminder to all of us - you, @miha-stopar, @zitnik and myself - since I believe none of us used `dep` workflows before).

Makefile
janhartman: Hm, yes, I worded it a bit weirdly - I meant "To populate ... and potentially modify" 😄

PR #75, opened 2018-03-05
New structs for pseudonym system keys

mancabizjak:
This PR introduces three new structs: `PubKey`, `PubKeyEC` and `SecKey` that represent  public keys for both implementations of pseudonym system scheme, and their secret keys, which are of the same type
for both implementations. `PubKey` and `PubKeyEC` are not only a different naming for our previous `OrgPubKeys` and `OrgPubKeysEC` types (these were removed), but are more generic, since CA's also have their corresponding `PubKey`.

Several structs and functions from the `pseudonymsys` package previously accepted parameters in the form of: `s1, s2 *big.Int`, `h1, h2 *groups.ECGroupElement` or similar - this PR modifies such structs and functions so that they accept appropriate key struct types.

Packages `client,` `compatibility`, `server` and `config` were updated in light of these changes.

I believe these new structs make the code a bit easier to follow for the readers unfamiliar with cryptographic core of emmy, since you don't have to be know, for instance, that `s1,s2` or `x,y` represent parts of a secret key, because this PR makes it obvious by introducing appropriate struct types.

@zitnik please review the changes to `compatibility` package. I can modify names of wrapper types if you prefer.

### DIFF
--- a/client/compatibility/doc.go
+++ b/client/compatibility/doc.go
@@ -65,8 +65,8 @@
 //	Credential
 //	CredentialEC
 //	ECGroupElement
-//	OrgPubKeys
-//	OrgPubKeysEC
+//	PubKey
+//	PubKeyEC
 //	Transcript
 //	TranscriptEC
 // 	Pseudonym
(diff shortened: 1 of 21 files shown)


mancabizjak:
Done.

PR #79, opened 2018-03-26
Remove communication for demo protocols

mancabizjak:
## Purpose
We agreed offline to remove communication layer for simple protocols that were initially included in emmy for demo purposes, as they complicate necessary refactorings that are common at this stage. 

## Scope
This PR cleans up definitions in the `protobuf` package:
* `Protocol` service (dedicated to running interactive protocols for demo purposes) is removed.
* All message types that are not used in pseudonym system scheme or in QR proofs are removed.
* A new service called `QRProofs` with RPCs `QR` and `QNR` is added.
* Enumerations are removed, because there will be no need to differentiate protocols with respect to `SchemaType` and `SchemaVariant` (they were only used in the protocols that were removed in this PR).

In addition, it removes all the communication code of the protocols included for demonstration purposes (_schnorr_, _schnorr_ec_, _pedersen_, _pedersen_ec_ and _cspaillier_). The only thing we wanted to keep communication-wise were qr proofs, which are now a part of a separate RPC service, so some changes were required in QR/QNR client and server code.

All the client subcommands of the emmy CLI corresponding to demo protocols were removed. A deprecation notice is added to the instructions of the emmy client in the README.


mancabizjak:
@miha-stopar As discussed yesterday, I also removed QR and QNR. I got myself into a bit of a merge hell, so I had to reset history. Everything is in a single commit now - all that has changed from yesterday is the removal of QR and QNR client-server code + removal of appropriate potobuffer service and message types.

PR #84, opened 2018-04-03
Schnorr refactoring

miha-stopar:
The biggest change is the unification of RepresentationProver and SchnorrProver. Until now SchnorrProver and RepresentationProver were two different entities, now there is only a SchnorrProver which offers a generalization previously implemented in RepresentationProver.

Also, ZKP and ZKPOK options have been removed - a different technique to transform Sigma to ZKP will be implemented later.

Pedersen commitment has been slightly changed - there were two methods doing the same thing.


miha-stopar:
Thanks, Manca. I think I addressed all of the comments except a few where some parameters depend on the others and have to instantiated before calls like "return &PedersenReceiver".

mancabizjak:
Oops, you're right, my bad. Anyways, looks good to me :+1:

PR #85, opened 2018-04-05
Representation proof for QRSpecialRSA implemented.

miha-stopar:
Representation proof is like Schnorr proof but in a QRSpecialRSA group instead of Schnorr group.

Signed-off-by: miha <stopar.miha@gmail.com>

### DIFF
--- a/crypto/zkp/primitives/qrspecialrsaproofs/representation.go
+++ b/crypto/zkp/primitives/qrspecialrsaproofs/representation.go
@@ -25,11 +25,11 @@ import (
 )
 
 // RepresentationProver is like SchnorrProver but in a QRSpecialRSA group (note that here proof data is
-// is computed in Z, not modulo as in Schnorr). Also, RepresentationProver with only one base and one secret
+// computed in Z, not modulo as in Schnorr). Also, RepresentationProver with only one base and one secret
 // is very similar to the DFCommitmentOpeningProver (RepresentationProver does not have a committer though).
 type RepresentationProver struct {
 	group      *groups.QRSpecialRSA
-	l_r        int // security parameter
+	secParam   int // security parameter
 	secrets    []*big.Int
 	bases      []*big.Int
 	randomVals []*big.Int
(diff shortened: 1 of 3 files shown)


crypto/zkp/primitives/qrspecialrsaproofs/representation.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"math/big"
+
+	"github.com/xlab-si/emmy/crypto/common"
+	"github.com/xlab-si/emmy/crypto/groups"
+)
+
+// RepresentationProver is like SchnorrProver but in a QRSpecialRSA group (note that here proof data is
+// is computed in Z, not modulo as in Schnorr). Also, RepresentationProver with only one base and one secret
```
mancabizjak: One "is" too many

crypto/zkp/primitives/qrspecialrsaproofs/representation.go
```diff
@@ -0,0 +1,125 @@
...
+// is very similar to the DFCommitmentOpeningProver (RepresentationProver does not have a committer though).
+type RepresentationProver struct {
+	group      *groups.QRSpecialRSA
+	l_r        int // security parameter
```
mancabizjak: Could you please rename this variable to either something more descriptive (for instance, `secParam`), or maybe `k` that is commonly used to denote security parameter - something that does not require underscores? The convention is to use mixedCaps.

This comment applies to all usages of `l_r`.

crypto/zkp/primitives/qrspecialrsaproofs/representation_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+// TestQRSpecialRSA demonstrates how to prove that for a given y you know secrets x_1,...,x_k
+// such that y = g_1^x_1 * ... * g_k^x_k where g_i are given generators (bases) of QRSpecialRSA.
+func TestQRSpecialRSA(t *testing.T) {
+	group, err := groups.NewQRSpecialRSA(1024)
+	if err != nil {
+		t.Errorf("error when creating QRSpecialRSA group: %v", err)
+	}
+
+	var bases [3]*big.Int
```
mancabizjak: Please modify L37-L38 like this:

```go
bases := make([]*big.Int, 3)
	for i := 0; i < len(bases); i++ { ... }
```
There's no need to first define an array and use `bases[:]` as arguments to `NewRepresentationProver` and `SetProofRandomData`. I think it's better to just define a slice at the beginning and then simply pass `bases` to these functions.

The same holds for `secrets`.

miha-stopar:
Thanks! Comments addressed.

PR #88, opened 2018-05-14
Camenisch-Lysyanskaya credential issuance 

miha-stopar:
First version of CL credential issuance - still missing hidden and committed attributes.


crypto/zkp/schemes/cl/org_credential_issuer.go
```diff
@@ -0,0 +1,176 @@
...
+			return false
+		}
+	}
+	if UProofData[len(i.Org.PubKey.RsKnown)].Cmp(b2) > 0 {
+		return false
+	}
+	return true
+}
+
+func (v *OrgCredentialIssuer) VerifyCredentialRequest(nymProof *dlogproofs.SchnorrProof,
```
mancabizjak: Rename method receiver to `i` for consistency.

PR #89, opened 2018-05-29
Credential issue attributes

miha-stopar:
First version of Camenisch-Lysyanskaya credential issuance implemented.


crypto/zkp/schemes/cl/user.go
```diff
@@ -26,22 +26,45 @@ import (
...
+		PubKey:             clPubKey,
+		Committers:         make(map[string]*commitments.PedersenCommitter),
+		PedersenParams:     pedersenParams,
+		knownAttrs:         knownAttrs,
+		committedAttrs:     committedAttrs,
+		hiddenAttrs:        hiddenAttrs,
+		commitmentsOfAttrs: commitmentsOfAttrs,
+		attrsCommitters:    attrsCommitters,
+	}, nil
 }
```
mancabizjak: Again, something that I probably should have thought of in the last review, but nevertheless: why do we let `nymName` to be passed to  `User`'s `GenerateNym` method from the outside? What happens if `GenerateNym` is called twice with the same `nymName`? Would it not be better to have a `GenerateNym` method that takes no arguments, and instead generates some unique random string itself as the nym name?

In addition, if `Nym`s are now associated with names, maybe `GenerateNym` could return a `Nym` or `NamedNym` struct that would hold both the value and the corresponding name.

As it is now, allowing `nymName` to be passed from the outside seems a bit off to me.

mancabizjak: Since `GenerateMasterSecret` doesn't return any value, but merely sets the `masterSecret` field, I propose that you:
1. Copy L71 to and move initialization of the `masterSecret` field to `NewUser` function. 
1. Remove the `GenerateMasterSecret` method.

Why? Well, what would happen if I created a `User` but forgot to explicitly call its `GenerateMasterSecret` method? The `masterSecret` field would be `nil`, and thus the call to `GenerateNym` would panic. This could never happen with the proposed change.

crypto/zkp/schemes/cl/user_credential_receiver.go
```diff
@@ -137,24 +145,45 @@ func (r *UserCredentialReceiver) getCredentialRequestProoRandomfData(nymName str
...
+			proofData1, proofData2)
+	}
+	return commitmentsOfAttrsProofs
 }
 
-func (r *UserCredentialReceiver) GetCredentialRequest(nymName string, nym, U, n1 *big.Int) (*dlogproofs.SchnorrProof,
-	*qrspecialrsaproofs.RepresentationProof, error) {
-	nymProofRandomData, UProofRandomData, err := r.getCredentialRequestProoRandomfData(nymName)
+// GetCredentialRequest computes U ...
+func (r *UserCredentialReceiver) GetCredentialRequest(nymName string, nym, n1 *big.Int) (*dlogproofs.SchnorrProof,
```
mancabizjak: 1. See one of the other comments where I propose encapsulating nym value + name into a single struct called `Nym` or `NamedNym`. As it is now, we could just pass any nym name that does not correspond to the nym value at all, and it could lead to unexpected behavior.
2. From the test code, it is obvious that parameter `n1` (and in some other functions also `n2`) corresponds to nonce. For clarity, could you please **rename all nonce parameters - here and in other functions** - to `nonce`? It would add some meaning to the parameter, whereas `n1` and `n2` hold no meaning to an average Joe at all :slightly_smiling_face:

crypto/zkp/schemes/cl/user.go
```diff
@@ -26,22 +26,45 @@ import (
...
 
-func NewUser(clParamSizes *CLParamSizes, clPubKey *PubKey, pedersenParams *commitments.PedersenParams) *User {
-	return &User{
-		ParamSizes:     clParamSizes,
-		PubKey:         clPubKey,
-		Committers:     make(map[string]*commitments.PedersenCommitter),
-		PedersenParams: pedersenParams,
-		attrs:          []*big.Int{big.NewInt(7), big.NewInt(6), big.NewInt(5)}, // TODO attributes should be read from somewhere and the structure should be loaded too
+func NewUser(clParamSizes *CLParamSizes, clPubKey *PubKey, pedersenParams *commitments.PedersenParams,
```
mancabizjak: A question - are the values of user's `CLParamSizes` in any way connected to known/committed/hidden attributes?

I was thinking that, lets say, I set some values of `CLParamSizes`, but then pass some attributes which do not conform to some of these values from `CLParamSizes`. Can this happen? If it can, we need to introduce some checks to see if parameter values are consistent with the rest of arguments we're getting in the `NewUser` function.

miha-stopar:
Thanks, Manca. Yes, we are getting there! 

Good points! I pretty much fixed everything according to your comments. For credential struct - I plan to add it when implementing update credential functionality. For nym identifiers - something I still need to think a bit, but I fully agree.

PR #91, opened 2018-06-11
Credential update implemented

miha-stopar:
Signed-off-by: Miha Stopar <miha.stopar@xlab.si>

### DIFF
--- a/crypto/zkp/schemes/cl/credential_issue_test.go
+++ b/crypto/zkp/schemes/cl/credential_issue_test.go
@@ -82,8 +82,8 @@ func TestCLIssue(t *testing.T) {
 
 	nonceUser1 := userCredentialReceiver.GetNonce()
 	newKnownAttrs := []*big.Int{big.NewInt(17), big.NewInt(18), big.NewInt(19), big.NewInt(27)}
+	user.UpdateCredential(newKnownAttrs)
 	credential1, AProof1 := orgCredentialIssuer.UpdateCredential(nonceUser1, newKnownAttrs)
-	user.knownAttrs = newKnownAttrs
 
 	userVerified, err = userCredentialReceiver.VerifyCredential(credential1, AProof1, nonceUser1)
 	if err != nil {
--- a/crypto/zkp/schemes/cl/credential_issue_test.go
+++ b/crypto/zkp/schemes/cl/credential_issue_test.go
@@ -49,6 +49,7 @@ func TestCLIssue(t *testing.T) {
 		t.Errorf("error when generating nym: %v", err)
 	}
 
+	// TODO: change API - move issuer into org, receiver into user
 	orgCredentialIssuer, err := NewOrgCredentialIssuer(org, nym.Commitment, user.knownAttrs,
 		user.commitmentsOfAttrs)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


crypto/zkp/schemes/cl/credential_issue_test.go
```diff
@@ -67,17 +67,28 @@ func TestCLIssue(t *testing.T) {
...
 	if err != nil {
 		t.Errorf("error when verifying credential: %v", err)
 	}
 
 	assert.Equal(t, true, userVerified, "credential issuance failed")
+
+	nonceUser1 := userCredentialReceiver.GetNonce()
+	newKnownAttrs := []*big.Int{big.NewInt(17), big.NewInt(18), big.NewInt(19), big.NewInt(27)}
+	credential1, AProof1 := orgCredentialIssuer.UpdateCredential(nonceUser1, newKnownAttrs)
+	user.knownAttrs = newKnownAttrs
```
mancabizjak: Setting of user's `knownAttrs` should be "hidden away", or done somehow differently, IMO. Do you think it would make sense to add a `UpdateCredential` method on `UserCredentialReceiver` or `User`, just so that it's clear both sides need to update this value with a function call?

I tried commenting out L86 and the test failed, but to me it was not immediately obvious why. I see now that `UserCredentialReceiver` takes a pointer to `User`. It's a bit confusing, but I believe a pending TODO at L52 (concerning moving `UserCredentialReceiver` into `User`) can improve this. Still, perhaps adding `UpdateCredential` now isn't a bad idea.

crypto/zkp/schemes/cl/org_credential_issuer.go
```diff
@@ -153,20 +148,17 @@ func (i *OrgCredentialIssuer) verifyCommitmentsOfAttrs(commitmentsOfAttrsProofs
 	return true
 }
 
-func (i *OrgCredentialIssuer) VerifyCredentialRequest(nymProof *dlogproofs.SchnorrProof, U *big.Int,
-	UProof *qrspecialrsaproofs.RepresentationProof,
-	commitmentsOfAttrsProofs []*commitmentzkp.DFOpeningProof) bool {
-	i.U = U
+func (i *OrgCredentialIssuer) VerifyCredentialRequest(credentialRequest *CredentialRequest) bool {
+	i.U = credentialRequest.U
```
mancabizjak: Please use shorter parameter names, for instance replace `credentialRequest` with `req`, or even better `cr`, because it's the way to Go (:smile: ). Otherwise it reads too much like Java :)

crypto/zkp/schemes/cl/user_credential_receiver.go
```diff
@@ -165,53 +165,69 @@ func (r *UserCredentialReceiver) getCommitmentsOfAttrsProof(challenge *big.Int)
...
+		commitmentsOfAttrsProofs), nil
 }
 
 func (r *UserCredentialReceiver) GetNonce() *big.Int {
 	b := new(big.Int).Exp(big.NewInt(2), big.NewInt(int64(r.User.Params.SecParam)), nil)
 	return common.GetRandomInt(b)
 }
 
-func (r *UserCredentialReceiver) VerifyCredential(A, e, v11 *big.Int,
+func (r *UserCredentialReceiver) VerifyCredential(credential *Credential,
```
mancabizjak: Please use `cred` or maybe `c` as  a parameter name instead of `credential`. Go favors shorter parameter names,  especially for variables with small scopes.

miha-stopar:
Thanks, Manca. Agreed for user.UpdateCredential. There are actually many things to be improved in the API, which I plan to do in the next PRs.

PR #94, opened 2018-07-27
Safe prime Schnorr group removed.

miha-stopar:
Signed-off-by: Miha Stopar <miha.stopar@xlab.si>


miha-stopar:
SchnorrSafePrimeGroup removed because it is not needed (r in p = q*r + 1 should be much bigger than 2 to obtain speed-up in performance).

PR #95, opened 2018-08-01
CL scheme communication layer

miha-stopar:
CL scheme API refactored. Communication layer for CL scheme implemented. Also, Redis is used for storing credential records on the server side - any suggestions how to modify DB layer are welcome, it is just a prototype.


README.md
```diff
@@ -22,15 +21,153 @@ various parts of Emmy library, please refer to additional documentation in the *
...
+Now a student has a credential which can be used 
+
+
+
+ 
+### How does CL work
+ 
+ 
+
+# TODO
```
mancabizjak: 1. I  don't like some of the bullets. If we want to tell the community what are our plans for the future, we need to be more precise, and focus on the what needs to be done in terms of core functionality. For instance, saying "Fix Camenisch-Shoup verifiable encryption" makes me wonder that it isn't working and therefore shouldn't be offered in the first place. Also, stating "Refactor proofs to set all parameters in a constructor" and direct the reader to a comment in the code feels wrong. This is probably more relevant for us as developers, let's use trello for that :wink:. Also, "Performance optimization" of which parts, can you be more specific?
2. This TODO list should be the last thing in the README.
3. Maybe it would be better to call this TODO list _Roadmap_

README.md
```diff
@@ -22,15 +21,153 @@ various parts of Emmy library, please refer to additional documentation in the *
...
+ * Additional proofs for Camenisch-Lysyanskaya scheme (range proof for attributes ... )
+ * Revocation for Camenisch-Lysyanskaya scheme
+ * Attribute types in Camenisch-Lysyanskaya scheme (string, int, date, enum)
+ * Performance optimization
+ * Efficient attributes for anonymous credentials [15]
+ * Camenisch-Lysyanskaya scheme based on pairings [14]
+ * Fix Camenisch-Shoup verifiable encryption
+ * Refactor proofs to set all parameters in a constructor (see comment in Schnorr prover)
+ 
 # Installation
```
mancabizjak: Can you please add this before Installation section (and also update the TOC):

# Warning
_All components of Emmy cryptography library are a work in progress. At this point, the library can be used to build proof of concept implementations for research purposes and **should never be used in production**. Project's code organization and library APIs are **not stable** - they are expected to undergo major changes, and may be changed at any point._

PR #99, opened 2018-08-30
Retire crypto/zkp/commitments, crypto/commitments

mancabizjak:
Continues refactoring series (#97, #98). 

This PR retires `crypto/zkp/commitments` and `crypto/commitments` packages. Commitments were placed into new packages  `df`, `qoneway`, `pedersen` and `ecpedersen` packages.

I also updated the README and docs/theory document.


mancabizjak:
I just noticed we don't mention preimage proofs anywhere in the README. Should I add an entry under the _Zero-knowledge proofs_ section?

PR #104, opened 2018-09-20
Move pseudonym system schemes

mancabizjak:
This PR moves core of the pseudonym system scheme to two separate packages - `crypto/pseudys` and `crypto/ecpseudsys`. This leaves our old package `crypto/zkp/schemes` empty, so it was removed.

Just like in #101, many of the struct names and names of their fields  (for instance `Credential` -> `Cred`, `Pseudonym` -> `Nym`, `Certificate` -> `Cert`,  field `equalityVerifier` -> `verifier` etc.) were shortened. In addition, several struct fields are now unexported, as there seemed to be no need for exporting them.

I kept struct names in `client/compatibility` package as they were (we'll focus on that later).


client/pseudonymsys_ec.go
```diff
@@ -51,18 +51,18 @@ func (c *PseudonymsysClientEC) GenerateMasterKey() *big.Int {
...
+	*ecpseudsys.Nym, error) {
 	if err := c.openStream(c.grpcClient, "GenerateNym_EC"); err != nil {
 		return nil, err
 	}
 	defer c.closeStream()
 
 	prover := ecschnorr.NewEqualityProver(c.curve)
 
-	// Differently as in Pseudonym Systems paper a user here generates a nym (if master
+	// Differently as in Nym Systems paper a user here generates a nym (if master
```
miha-stopar: Shoud be Pseudonym System paper - that's how paper is named.

mancabizjak: Oops :) Done.

PR #110, opened 2018-11-21
Verifiable encryption of elliptic curve secrets

tetratorus:
Implement functions and tests for verifiable encryption of a scalar secret on the secp256k1 curve.


miha-stopar:
Many thanks for PR, @tetratorus! It seems ok to me, however there is one problem. In the last months we have rewritten pretty much every primitive and scheme except the one for verifiable encryption. We plan to do it, but not sure if in the near future. It badly needs refactoring, however we have no need for this primitive at the moment.

If there is an application where you would like to use verifiable encryption, we can perhaps do this sooner than initially planned. Bottom line - I would merge the test you provided, but would wait with elliptic curve group until we refactor the primitive (it would not require copying the whole file then).

miha-stopar:
I would then suggest you modify PR to include only the extension of a test and fix related to keys. I will then merge it. Also, can you please delete one PR? It seems like there are two pretty much the same PRs.

PR #112, opened 2019-01-18
CL API

miha-stopar:
Camenisch-Lysyanskaya anonymous credentials scheme has now slightly more intuitive API for handling attributes. Also, two gRPC services have been added - to obtain credential structure required by an organization; and to obtain list of acceptable credentials.


crypto/cl/raw_creds.go
```diff
@@ -0,0 +1,138 @@
...
+		attrNameToIndex: make(map[string]int),
+	}
+}
+
+func (c *RawCredential) InsertAttribute(index int, name, attrType string, known bool) {
+	c.attrNameToIndex[name] = index
+	c.attributes[index] = NewAttribute(index, name, attrType, known, nil)
+}
+
+func (c *RawCredential) AddAttribute(name, attrType string, known bool, value string) error {
```
mancabizjak: I think we could further simplify `AddAttribute`.

If we support int and string attributes, I think it's a good idea to offer `AddStringAttribute` and `AddIntAttribute`. As an end-user app, I would love to be able to do something like this:

````go
_ := c.AddStringAttribute("Name", "myname") // for a known string attribute
_ := c.AddIntAttribute("Age", 30).Hide() // for a hidden int attribute
````
The above example assumes that all attributes are known by default, and we can control their visibility with a method like `Hide()`. This way, we're left with an even more intuitive API, since we only need to provide attribute name and value.

File: crypto/cl/raw_creds.go [hunk 2]
```diff
@@ -0,0 +1,138 @@
...
+	index := len(c.attributes)
+	c.attrNameToIndex[name] = index
+	c.attributes[index] = NewAttribute(index, name, attrType, known, nil)
+	return c.SetAttributeValue(name, value)
+}
+
+// SetAttributeValue converts and set the attribute value to *big.Int which is then used internally
+// by CL scheme.
+func (c *RawCredential) SetAttributeValue(name string, value string) error {
```
mancabizjak: I wonder if this should be a method of `Attribute` rather than `RawCredential`. To me it makes sense that we have `AddAttribute` method on the `RawCredential`, but every supported attribute type (int or string) should know how to convert its original value to a `*big.Int` representation by itself - I think this is not `RawCredential`'s responsibility.

I propose having a `func (a *Attribute) InternalValue()` that would perform the conversion. Of course, if we decide to split attributes in `IntAttribute`s and `StringAttribute`s, we'd need separate implementations of this method. So `InternalValue()` could be an interface method.

crypto/cl/raw_creds.go
```diff
@@ -0,0 +1,138 @@
...
+	index := c.attrNameToIndex[name]
```
mancabizjak:
```suggestion
	i := c.attrNameToIndex[name]
```

crypto/cl/raw_creds.go
```diff
@@ -0,0 +1,138 @@
...
+		Name:  name,
+		Type:  attrType,
+		Known: known,
+		Value: value,
+	}
+}
+
+// RawCredential presents credential as to be used by application that executes CL scheme to prove possesion
+// of an anonymous credential.
+type RawCredential struct {
```
mancabizjak:
```suggestion
type RawCred struct {
```

crypto/cl/raw_creds.go
```diff
@@ -0,0 +1,138 @@
...
+	attributes      map[int]*Attribute
+	attrNameToIndex map[string]int
+}
+
+func NewRawCredential() *RawCredential {
```
mancabizjak:
```suggestion
func NewRawCred() *RawCred {
```

proto/services.proto
```diff
@@ -36,6 +36,11 @@ service PseudonymSystem {
 	rpc TransferCredential_EC (stream Message) returns (stream Message) {}
 }
 
+service CLCredentialInfo {
+	rpc GetCredentialStructure(google.protobuf.Empty) returns (CredentialStructure) {}
```
mancabizjak:
```suggestion
	rpc GetCredStructure(google.protobuf.Empty) returns (CredStructure) {}
```

proto/services.proto
```diff
@@ -36,6 +36,11 @@ service PseudonymSystem {
...
+	rpc GetAcceptableCredentials(google.protobuf.Empty) returns (AcceptableCredentials) {}
```
mancabizjak:
```suggestion
	rpc GetAcceptableCreds(google.protobuf.Empty) returns (AcceptableCreds) {}
```

server/cl_info.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func (s *Server) GetAcceptableCredentials(ctx context.Context, _ *empty.Empty) (*pb.AcceptableCredentials, error) {
+	s.Logger.Info("Client requested acceptable credentials information")
+	accCreds, err := config.LoadAcceptableCredentials()
+	if err != nil {
+		return nil, err
+	}
+
+	var credentials []*pb.AcceptableCred
```
mancabizjak:
```suggestion
	var creds []*pb.AcceptableCred
```

proto/services.proto
```diff
@@ -36,6 +36,11 @@ service PseudonymSystem {
 	rpc TransferCredential_EC (stream Message) returns (stream Message) {}
 }
 
+service CLCredentialInfo {
```
mancabizjak: I think this service unnecessary, maybe we can just add these two RPCs to the existing CL service.

client/cl.go
```diff
@@ -126,14 +191,34 @@ func (c *CLClient) UpdateCredential(credManager *cl.CredManager, newKnownAttrs [
 	return nil, fmt.Errorf("credential not valid")
 }
 
-// ProveCred proves the possession of a valid credential and reveals only the attributes the user desires
+// ProveCredential proves the possession of a valid credential and reveals only the attributes the user desires
 // to reveal. Which knownAttrs and commitmentsOfAttrs are to be revealed are given by revealedKnownAttrsIndices and
 // revealedCommitmentsOfAttrsIndices parameters. All knownAttrs and commitmentsOfAttrs should be passed into
-// ProveCred - only those which are revealed are then passed to the server.
+// ProveCredential - only those which are revealed are then passed to the server.
```
mancabizjak: Please update the comment.

config/defaults.yml
```diff
@@ -52,6 +52,18 @@ service_info:
   provider: "Government"
   description: "This service verifies your right to vote and allows you to vote electronically with cryptographically assured anonymity"
 
+# the number of attributes must correspond to the CL params (see KnownAttrsNum, 
+# CommittedAttrsNum, HiddenAttrsNum)
+attributes: {0: "Name, string, true", 1: "Gender, string, true", 2: "Graduated, string, true", 
+3: "DateMin, int64, true", 4: "DateMax, int64, true", 5: "Age, int64, false"}
+
+# credentials from which organizations are accepted and which attributes need to be revealed
+acceptable_credentials: {"Org1": "Name, DateMin, DateMax", "Org2": "Gender"}
```
mancabizjak: Same comment as for `attributes` in the `default.yml`

config/defaults.yml
```diff
@@ -52,6 +52,18 @@ service_info:
...
+#conditions: {3: "greater", 4: "lesser"}
+int_values: {3: 1562643000, 4: 1562643000}
```
mancabizjak: Same comment as for `attributes` in the `default.yml`

crypto/cl/encoder.go
```diff
@@ -0,0 +1,46 @@
...
+
+package cl
+
+import (
+	"encoding/gob"
+	"os"
+)
+
+// TODO: where should we put WriteGob and ReadGob?
+func WriteGob1(filePath string, object interface{}) error {
```
mancabizjak: Why does the function name have suffix `1`?

crypto/cl/encoder.go
```diff
@@ -0,0 +1,46 @@
...
+	if err == nil {
+		encoder := gob.NewEncoder(file)
+		encoder.Encode(object)
+	}
+	file.Close()
+
+	return err
+}
+
+func ReadGob1(filePath string, object interface{}) error {
```
mancabizjak: Why does the function name have suffix `1`?

crypto/cl/org.go
```diff
@@ -259,6 +260,74 @@ func (o *Org) GetProveCredNonce() *big.Int {
...
+		indexAll := revealedIndices[ind]
+		cond := conditions[indexAll]
+
+		switch val.(type) {
+		case int:
+			valInt := val.(int)
+			accVal := intValues[indexAll]
+
+			if cond == "greater" {
+				if accVal < valInt {
```
mancabizjak: No need to nest `if`s, you can just `cond == "greater" && accVal < valInt`

crypto/cl/org.go
```diff
@@ -259,6 +260,74 @@ func (o *Org) GetProveCredNonce() *big.Int {
...
+					return false, fmt.Errorf("attribute value for %s not acceptable", a.GetName())
+				}
+			} else if cond == "lesser" {
```
mancabizjak: No need to nest `if`s, just join the conditions.

crypto/cl/org.go
```diff
@@ -259,6 +260,74 @@ func (o *Org) GetProveCredNonce() *big.Int {
...
+				if accVal > valInt {
+					return false, fmt.Errorf("attribute value for %s not acceptable", a.GetName())
+				}
+			} else if cond == "same" {
```
mancabizjak: No need to nest `if`s, just join the conditions. Also, please use ` accVal != valInt` for consistency.

crypto/cl/org.go
```diff
@@ -259,6 +260,74 @@ func (o *Org) GetProveCredNonce() *big.Int {
...
+				if valInt != accVal {
+					return false, fmt.Errorf("attribute value for %s not acceptable", a.GetName())
+				}
+			}
+		case string:
+			accVal := strValues[indexAll]
+			if cond == "same" {
```
mancabizjak: Same comment as above, no need to nest `if`s.

crypto/cl/org.go
mancabizjak: Also, if we have `lesser` and `greater`, maybe it would be more intuitive to use `equal` instead of `same`.

README.md
```diff
@@ -31,6 +31,97 @@ Emmy library is named after a German mathematician [Emmy Noether](https://en.wik
 - [Further documentation](#documentation)
 <!-- tocstop -->
 
+# emmy - anonymous credentials showcase
```
mancabizjak: This section needs some updating, according to the changes from 9f25aa2b556a61063ed13990da3901b3fb7b8add

PR #114, opened 2020-12-04
small fix for README

stefan2904:
just a small fix. inline link to anchor did not work so I made it work. :)


miha-stopar:
Nice, but it should link to schemes, not primitives. If you fix this and another typo (search for cryptograhpic), I will merge it.

miha-stopar:
Nice, but it should link to schemes, not primitives. If you fix this and another typo (search for cryptograhpic), I will merge it.

